Issue #7, opened 2018-01-23
Lower case requirement + small GUI glitch


pako81:
When having the following password requirements:

- 9 minimum characters

- 1 uppercase letters

- 1 numbers

- 1 special characters and defining them as !@#$%^&*()_-+=`~,./?;:'"[{]}|

it will accept !@#$1234A as a password. We'd ideally like for it to enforce a lower case letter as well. Would this be possible to have added?

Another thing that I noticed was when the special characters field is empty, it says to separate the characters with space or comma (which from what I can tell, isn't required).

Tested this on 10.0.4 + Password Policy app 1.1.0.

DeepDiver1975:
> it says to separate the characters with space or comma (which from what I can tell, isn't required).

fixed already with #6

PR #8, opened 2018-01-24
Added Lowercase rule

mdusher:
Defined a minimum lowercase letters rule and tests, should cover #7.


lib/Rules/Lowercase.php
```diff
@@ -0,0 +1,45 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\PasswordPolicy\Rules;
+
+class lowercase extends Base {
```
DeepDiver1975: Should be Lowercase

mdusher:
Ah whoops! Fixed it!

Issue #3, opened 2018-01-22
Password history


DeepDiver1975:
Expectation:
- Passwords need to be different than the previous ones (history of, say, 10 previous passwords)
- history length is an app config vale
- list of hashes is stored in a new database table
- unit test need to include myqsl, postgres and oracle - automated on travis

DeepDiver1975:
fixed with #10

PR #10, opened 2018-01-24
add password history rule

karakayasemi:
implementation of #3.

### DIFF
--- a/appinfo/Migrations/Version20180123143343.php
+++ b/appinfo/Migrations/Version20180123143343.php
@@ -1,0 +1,58 @@
+<?php
+/**
+ *
+ * @author Semih Serhat Karakaya <karakayasemi@itu.edu.tr>
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license GPL-2.0
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+namespace OCA\password_policy\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use OCP\Migration\ISchemaMigration;
+
+class Version20180123143343 implements ISchemaMigration {
+
+	/** @var  string */
+	private $prefix;
+
+	public function changeSchema(Schema $schema, array $options) {
+		$this->prefix = $options['tablePrefix'];
+		if (!$schema->hasTable("{$this->prefix}user_password_history")) {
+			$table = $schema->createTable("{$this->prefix}user_password_history");
+			$table->addColumn('id', 'integer', [
+				'autoincrement' => true,
+				'unsigned' => true,
+				'notnull' => true,
+				'length' => 11,
+			]);
+			$table->addColumn('uid', 'string', [
+				'notnull' => true,
+				'length' => 64,
+			]);
+			$table->addColumn('password', 'text', [
+				'notnull' => true,
+				'length' => 255,
+			]);
+			$table->addColumn('change_time', 'integer', [
+				'notnull' => true,
+			]);
+
+			$table->setPrimaryKey(['id']);
+			$table->addIndex(['uid'], 'pp_uid_index');
+		}
+	}
+}
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -29,7 +29,10 @@ $eventDispatcher->addListener(
 	'OCP\User::validatePassword',
 	function($event) {
 		if ($event instanceof GenericEvent) {
-			HooksHandler::verifyPassword($event->getArguments()['password']);
+			HooksHandler::verifyPassword(
+				$event->getArguments()['password'],
+				$event->getArguments()['uid']
+			);
 		}
 	}
 );
@@ -50,4 +53,12 @@ $eventDispatcher->addListener(
 		}
 	}
 );
-
+\OC::$server->getEventDispatcher()->addListener(
+	'user.aftersetpassword',
+	function (GenericEvent $event) {
+		HooksHandler::saveOldPassword(
+			$event->getArgument('user'),
+			$event->getArgument('password')
+		);
+	}
+);
--- a/lib/Db/OldPasswordMapper.php
+++ b/lib/Db/OldPasswordMapper.php
@@ -1,0 +1,54 @@
+<?php
+/**
+ *
+ * @author Semih Serhat Karakaya <karakayasemi@itu.edu.tr>
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license GPL-2.0
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\PasswordPolicy\Db;
+
+use OCP\AppFramework\Db\Mapper;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\IDBConnection;
+
+class OldPasswordMapper extends Mapper {
+	public function __construct(IDBConnection $db) {
+		parent::__construct($db, 'user_password_history');
+	}
+
+	/**
+	 * @param string $uid
+	 * @param int $length
+	 * @return OldPassword[]
+	 */
+	public function getOldPasswords($uid, $length) {
+		/* @var IQueryBuilder $qb */
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('password')
+			->from('user_password_history')
+			->where($qb->expr()->eq('uid', $qb->createNamedParameter($uid)))
+			->orderBy('change_time', 'desc');
+		$result = $qb->execute();
+		$rows = $result->fetchAll();
+		$result->closeCursor();
+		$rows = array_slice($rows,0, $length);
+		return array_map(function ($row) {
+			return OldPassword::fromRow($row);
+		}, $rows);
+	}
+}
--- a/lib/Engine.php
+++ b/lib/Engine.php
@@ -21,11 +21,15 @@
 
 namespace OCA\PasswordPolicy;
 
+use OCA\PasswordPolicy\Db\OldPasswordMapper;
 use OCA\PasswordPolicy\Rules\Length;
 use OCA\PasswordPolicy\Rules\Numbers;
 use OCA\PasswordPolicy\Rules\Special;
 use OCA\PasswordPolicy\Rules\Uppercase;
+use OCA\PasswordPolicy\Rules\PasswordHistory;
+use OCP\IDBConnection;
 use OCP\IL10N;
+use OCP\Security\IHasher;
 use OCP\Security\ISecureRandom;
 
 class Engine {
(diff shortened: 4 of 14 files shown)


karakayasemi:
@DeepDiver1975 please review. Thanks.

appinfo/Migrations/Version20180123143343.php
```diff
@@ -0,0 +1,47 @@
...
+
+use Doctrine\DBAL\Schema\Schema;
+use OCP\Migration\ISchemaMigration;
+
+class Version20180123143343 implements ISchemaMigration {
+
+    /** @var  string */
+    private $prefix;
+
+	public function changeSchema(Schema $schema, array $options) {
```
DeepDiver1975: indent

appinfo/Migrations/Version20180123143343.php
```diff
@@ -0,0 +1,47 @@
...
+        $this->prefix = $options['tablePrefix'];
+        if (!$schema->hasTable("{$this->prefix}user_password_history")) {
```
DeepDiver1975: not too long for oracle? let's see ;-)

appinfo/Migrations/Version20180123143343.php
```diff
@@ -0,0 +1,47 @@
...
+            $table = $schema->createTable("{$this->prefix}user_password_history");
+            $table->addColumn('uid', 'string', [
+                'notnull' => true,
+                'length' => 64,
+            ]);
+            $table->addColumn('password', 'text', [
```
DeepDiver1975: stored as clear text? if so this is a no go - please store it hashed

appinfo/app.php
```diff
@@ -50,4 +50,21 @@ function($event) {
 		}
 	}
 );
-
+$eventDispatcher->addListener(
+    'OCP\User::postSetPassword',
+    function($event) {
+        if ($event instanceof GenericEvent) {
+            $event['password'] = HooksHandler::generatePassword();
```
DeepDiver1975: why is here a password generated

lib/Db/OldPasswordMapper.php
```diff
@@ -0,0 +1,51 @@
...
+    public function getOldPasswords($uid, $length) {
+        /* @var IQueryBuilder $qb */
+        $qb = $this->db->getQueryBuilder();
+        $qb->select('password')
+            ->from('user_password_history')
+            ->where($qb->expr()->eq('uid', $qb->createNamedParameter($uid)))
+            ->orderBy('change_time', 'desc');
+        $result = $qb->execute();
+        $rows = $result->fetchAll();
+        $result->closeCursor();
```
DeepDiver1975: 👍

tests/rules/passwordhistorytest.php
```diff
@@ -0,0 +1,103 @@
...
+		$l10n = $this->getMockBuilder('\OCP\IL10N')
+			->disableOriginalConstructor()->getMock();
+		$l10n
+			->expects($this->any())
+			->method('t')
+			->will($this->returnCallback(function($text, $parameters = array()) {
+				return vsprintf($text, $parameters);
+			}));
+		/** @var \OCP\IUserSession | \PHPUnit_Framework_MockObject_MockObject $userSession */
+		$userSession = $this->getMockBuilder('\OCP\IUserSession')
```
DeepDiver1975: IUserSession::class

appinfo/Migrations/Version20180123143343.php
```diff
@@ -0,0 +1,47 @@
+<?php
+/**
+ * @author Semih Serhat Karakaya <karakayasemi@itu.edu.tr>
+ * @copyright Copyright (c) 2018, ownCloud GmbH.
+ * @license AGPL-3.0
```
DeepDiver1975: please change to GPLv2

appinfo/Migrations/Version20180123143343.php
karakayasemi: AFAIK, it supports up to 30 characters.

appinfo/Migrations/Version20180123143343.php
karakayasemi: no need to worry, I am storing as hashed

appinfo/app.php
karakayasemi: forgot to delete this listener

karakayasemi:
@DeepDiver1975  I believe it is ready to review this time.

lib/Rules/PasswordHistory.php
```diff
@@ -0,0 +1,58 @@
...
+class PasswordHistory extends Base {
+
+	/** @var IUserSession */
+	protected $userSession;
+
+	/** @var OldPasswordMapper */
+	protected $mapper;
+
+	/**
+	 * @param IL10N $l10n
```
DeepDiver1975: phpdoc missing

lib/Rules/PasswordHistory.php
```diff
@@ -0,0 +1,58 @@
...
+	 */
+	public function __construct(IL10N $l10n, IUserSession $userSession, OldPasswordMapper $mapper) {
+		parent::__construct($l10n);
+		$this->userSession = $userSession;
+		$this->mapper = $mapper;
+	}
+
+	public function verify($password, $val) {
```
DeepDiver1975: phpdoc missing

lib/Rules/PasswordHistory.php
```diff
@@ -0,0 +1,58 @@
...
+		$oldPasswords = $this->mapper->getOldPasswords(
+			$this->userSession->getUser()->getUID(),
+			$val
+		);
+		foreach($oldPasswords as $oldPassword) {
+			if (\OC::$server->getHasher()->verify($password, $oldPassword->getPassword())) {
+				throw new \Exception(
```
DeepDiver1975: please add throws annotation to phpdoc of this method

lib/Rules/PasswordHistory.php
```diff
@@ -0,0 +1,58 @@
...
+	 */
+	public function __construct(IL10N $l10n, IUserSession $userSession, OldPasswordMapper $mapper) {
+		parent::__construct($l10n);
+		$this->userSession = $userSession;
+		$this->mapper = $mapper;
+	}
+
+	public function verify($password, $val) {
+		$oldPasswords = $this->mapper->getOldPasswords(
+			$this->userSession->getUser()->getUID(),
```
DeepDiver1975: getUser can return null

appinfo/Migrations/Version20180123143343.php
```diff
@@ -0,0 +1,50 @@
...
+
+class Version20180123143343 implements ISchemaMigration {
+
+	/** @var  string */
+	private $prefix;
+
+	public function changeSchema(Schema $schema, array $options) {
+		$this->prefix = $options['tablePrefix'];
+		if (!$schema->hasTable("{$this->prefix}user_password_history")) {
+			$table = $schema->createTable("{$this->prefix}user_password_history");
```
DeepDiver1975: no auto increment id column - afaik this is necessary for the mapper to operate properly

appinfo/app.php
```diff
@@ -50,4 +50,12 @@ function($event) {
 		}
 	}
 );
-
+\OC::$server->getEventDispatcher()->addListener(
+	'user.aftersetpassword',
```
DeepDiver1975: this was introduced with 10.0.??? - info.xml needs update accordingly

lib/Engine.php
```diff
@@ -96,6 +98,12 @@ public function verifyPassword($password) {
 			$r = new Special($this->l10n);
 			$r->verify($password, $val, $chars);
 		}
+		if ($this->yes('spv_password_history_checked')) {
+			$val = $this->configValues['spv_password_history_value'];
+			$dbMapper = new OldPasswordMapper(\OC::$server->getDatabaseConnection());
```
DeepDiver1975: inject db connection

lib/Engine.php
```diff
@@ -96,6 +98,12 @@ public function verifyPassword($password) {
...
+			$r = new PasswordHistory($this->l10n, \OC::$server->getUserSession(), $dbMapper);
```
DeepDiver1975: inject user session

tests/rules/passwordhistorytest.php
```diff
@@ -0,0 +1,108 @@
...
+		/** @var \OCP\IUserSession | \PHPUnit_Framework_MockObject_MockObject $userSession */
+		$userSession = $this->getMockBuilder(IUserSession::class)
+			->disableOriginalConstructor()->getMock();
+
+		/** @var \OCP\IUser | \PHPUnit_Framework_MockObject_MockObject $user */
+		$user = $this->getMockBuilder(IUser::class)
+			->disableOriginalConstructor()->getMock();
+
+		/** @var \OCA\PasswordPolicy\Db\OldPasswordMapper | \PHPUnit_Framework_MockObject_MockObject $mapper */
+		$mapper = $this->getMockBuilder('\OCA\PasswordPolicy\Db\OldPasswordMapper')
```
DeepDiver1975: ::class

tests/rules/passwordhistorytest.php
```diff
@@ -0,0 +1,108 @@
...
+			->disableOriginalConstructor()->getMock();
+		/** @var \OCA\PasswordPolicy\Db\OldPassword | \PHPUnit_Framework_MockObject_MockObject $entity1 */
+		$entity1 = $this->getMockBuilder('\OCA\PasswordPolicy\Db\OldPassword')
```
DeepDiver1975: ::class

tests/rules/passwordhistorytest.php
```diff
@@ -0,0 +1,108 @@
...
+			->disableOriginalConstructor()->getMock();
+		/** @var \OCA\PasswordPolicy\Db\OldPassword | \PHPUnit_Framework_MockObject_MockObject $entity2 */
+		$entity2 = $this->getMockBuilder('\OCA\PasswordPolicy\Db\OldPassword')
```
DeepDiver1975: ::class

lib/Rules/PasswordHistory.php
karakayasemi: You are right but have you any suggestion for that?  'OCP\User::validatePassword' and 'OCP\Share::validatePassword' events are triggering this function and they are only supplying a password as the event argument. How can I learn password belongs to which user? We may need to add the user to the event arguments in the core.

DeepDiver1975: > We may need to add the user to the event arguments in the core.

should be the case already - if not the event was design the wrong way

E.g. the password have be changed via the terminal - there is no user session in this case.

Please adopt the event in core if necessary.

Thanks for pointing this out!

karakayasemi: as far as I can see, we have 4 validatePassword event emitting calls in the core and it is not possible to add the user object to event arguments in the all of them. 

https://github.com/owncloud/core/blob/92b658a60e6c28ef9dd7594384180e0d73e96660/lib/private/User/User.php#L253 : we can add here, also maybe only adding here can be sufficient for our purpose.

https://github.com/owncloud/core/blob/92b658a60e6c28ef9dd7594384180e0d73e96660/lib/private/User/Manager.php#L333 since there is no user yet, we can add the only uid in here, but since it user creation event, we don't need to apply password history to the newly created user.

Also, in sharing case, should we apply password history rule? If yes, we should use the creator of share in the validatePassword event.
https://github.com/owncloud/core/blob/92b658a60e6c28ef9dd7594384180e0d73e96660/lib/private/Share20/Manager.php#L174 
https://github.com/owncloud/core/blob/92b658a60e6c28ef9dd7594384180e0d73e96660/lib/private/Share/Share.php#L2867

@DeepDiver1975 what are your thoughts? IMHO, to ensure compatibility with all events, somehow we can add the only uid string in the validatePassword events. If you agree that, I can create a PR in the core for that.

DeepDiver1975: good points ... once again.

Yes - adding the uid should be fine. For the share password we dont set the uid because it is not reasonable. If no uid is set it is also pointless to check for the password history.

karakayasemi:
I adapted the code to the new validatePassword event with user parameter. When the core PR https://github.com/owncloud/core/pull/30326 merged, I will update info.xml to correct min version.

karakayasemi:
https://github.com/owncloud/core/pull/30326 merged and backported to stable10. It will be in the next release, as far as I understand it is 10.0.7. If I update min version to 10.0.7 probably unit test will fail. What should I do about info.xml?

DeepDiver1975:
> owncloud/core#30326 merged and backported to stable10. It will be in the next release, as far as I understand it is 10.0.7. If I update min version to 10.0.7 probably unit test will fail. What should I do about info.xml?

yes - 10.0.7 will be the minimum version we require. We can only set this once it has been release - please open an issue so that we don't forget this.

individual-it:
tested the feature with `occ`, changing own password in personal settings and as admin changing the password of a user, works as expected 👍

karakayasemi:
@DeepDiver1975 I need one more review for this PR, I hope it will be final.

lib/Rules/PasswordHistory.php
```diff
@@ -0,0 +1,65 @@
...
+		$this->hasher = $hasher;
+	}
+
+	/**
+	 * @param string $password
+	 * @param int $val
+	 * @param string $uid
+	 * @throws \Exception
+	 */
+	public function verify($password, $val, $uid) {
```
DeepDiver1975: if uid is null just return here - this check is only reasonable if the uid is given

karakayasemi: I checked the uid is not empty in the engine, but also 
added an additional check to here.

lib/Engine.php
```diff
@@ -96,6 +125,12 @@ public function verifyPassword($password) {
 			$r = new Special($this->l10n);
 			$r->verify($password, $val, $chars);
 		}
+		if ($this->yes('spv_password_history_checked') && !empty($uid)) {
```
DeepDiver1975: ah there is is .... ;-)

lib/HooksHandler.php
```diff
@@ -110,4 +119,17 @@ private static function loadConfiguration() {
 		}
 		return $configValues;
 	}
+
+	/**
+	 * @param \OC\User\User $user
```
DeepDiver1975: OCP\IUser would be better

PVince81:
10.0.7 will be another hotfix, any concerns or should min-version be set to 10.0.8 now ? @DeepDiver1975

DeepDiver1975:
We have an open issue to adjust the min version on release of the app

PR #12, opened 2018-02-26
Password policy is 10.1 ready

DeepDiver1975:


.travis.yml
```diff
@@ -42,6 +42,10 @@ matrix:
     - php: 5.6
       env: DB=mysql
     - php: 5.6
-      env: DB=mysql;TC=lint
+      env: DB=mysql TC=lint
+    - php: 5.6
+      env: DB=mysql CORE_BRANCH=stable10
```
phil-davis: since ``CORE_BRANCH=stable10`` already in ``global`` then this matrix entry will duplicate the one 2 entries above.

.travis.yml
```diff
@@ -42,6 +42,10 @@ matrix:
...
+    - php: 5.6
+      env: DB=mysql TC=lint CORE_BRANCH=stable10
```
phil-davis: ``CORE_BRANCH=stable10`` is redundant here.
Do you intend to run any matrix entries against core ``master`` ?

.travis.yml
DeepDiver1975: thanks for noticing - fixed - THX

Issue #2, opened 2018-01-22
Password expiry


DeepDiver1975:
Expectation:
- User passwords expire after X days
- User should be notified Y days before the password expires to change the password
- User should be notified when the password has expired

To be clarified:
- where to implement the password change dialog
  - [ ] core
  - [ ] this app

karakayasemi:
I've been working on the subject a little bit. In the user.afterlogin event we can check for expiry and send notification or e-mail to notify user x days before. However, I could not find any way to redirect a user to password change page from the app. I have two ideas to overcome this situation:
1. We can send an exception which gives the link to the password change page. Password change template can be stored in the app, in this way, theoretically, we don't need any change in the core. But, the user needs to click a button to go to password change page.
2. We can implement something similar with two-factor structure in the core, but it sounds a bit dirty and makes login structure even more complicated.

Shortly, the question is how can I redirect a user to password change page from an app? I'm not sure about my ideas. I would appreciate any help. Thanks.

DeepDiver1975:
My guts feeling is telling me the password change page should be part of core.
Summoning @PVince81 @butonic @tomneedham for more input ....

PVince81:
Ideally we should reuse and enhance the password change page from core.

Currently I think that page is the "password forgotten" page that is triggered by a "password forgotten" token routed there. Maybe a quick look at the guests app might help find out how, and see what link is sent in the email and how it is generated. I suspect that to reuse the current mechanism the password_policy app would also need to automatically generate this "password forgotten" token once before redirecting.

PVince81:
This was now implemented as an account module in the password policy app.

Remaining work for notifying about expiration was raised here: https://github.com/owncloud/password_policy/issues/18

PR #15, opened 2018-05-28
password expiry using an IAccountModule

butonic:
see core https://github.com/owncloud/core/pull/31547

- [x] password change template needs css polish
- [x] User needs to change password on first login -> listen on firstLogin event
- [x] change password dialog should start with cursor in old password field
- [x] template should compare new passwords in js without submitting
- [x] make redirect url work properly
- [x] check TODOs
- [x] tests
  - leaving HooksHandler fixDI & loadConfiguration untested, would require db access

left for a separate PR:
- [ ] notify users
  - [ ] once n days before expiry
  - [ ] every last k days before expiry
  - [ ] when password expired
  - [ ] should include password reset link
- [ ] additional AccountModule for failed login attempts

password expiry is checked on login. only affects current session so existing sessions are only affected if they do a login


lib/Controller/PasswordController.php
```diff
@@ -0,0 +1,159 @@
...
+	 * @param IRequest $request
+	 * @param IUserSession $userSession
+	 * @param IUserManager $userManager
+	 * @param IConfig $config
+	 * @param ISession $session
+	 * @param IURLGenerator $urlGenerator
+	 */
+	public function __construct($appName, IRequest $request, IUserSession $userSession, IUserManager $userManager, IConfig $config, ISession $session, IURLGenerator $urlGenerator) {
+		parent::__construct($appName, $request);
+		$this->user = $userSession->getUser(); // might be null
```
DeepDiver1975: no - get the user in the controller method - the user is set to the session ins some cases in the sec middleware

lib/Command/ExpirePassword.php
```diff
@@ -0,0 +1,87 @@
...
+	 */
+	public function __construct(IConfig $config, IUserManager $userManager) {
+		$this->config = $config;
+		$this->userManager = $userManager;
+		parent::__construct();
+	}
+
+	protected function configure() {
+		$this
+			->setName('user:expirepassword')
```
DeepDiver1975: expire-password

lib/Controller/PasswordController.php
```diff
@@ -0,0 +1,159 @@
...
+	 * Shows a change password template
+	 *
+	 * @param string $current_password
+	 * @param string $new_password
+	 * @param string $retyped_password
+	 * @param string $redirect_url
+	 * @return TemplateResponse|RedirectResponse
+	 *
+	 * @PublicPage
+	 * @NoCSRFRequired
```
DeepDiver1975: csrf shall be enabled for anything which is non-GET

lib/Db/OldPasswordMapper.php
```diff
@@ -1,54 +1,54 @@
-<?php
```
DeepDiver1975: what happened to this file?

lib/Rules/PasswordExpired.php
```diff
@@ -0,0 +1,89 @@
...
+			// TODO log warning? so the admin can force a change?
+			return;
+		}
+
+		$date = new \DateTime("@$changed");
+		$date->add(new \DateInterval("P{$days}D"));
+		$date->setTime(23, 59, 59); // include last day
+		$expiresAt = $date->getTimestamp();
+
+		$now = \OC::$server->getTimeFactory()->getTime();
```
DeepDiver1975: inject

lib/Db/OldPasswordMapper.php
butonic: wrong newlines

lib/Command/ExpirePassword.php
butonic: I wanted to be consistent with `user:resetpassword`, but whatever...

DeepDiver1975: should have been reset-password anyway ;-)

lib/Controller/PasswordController.php
butonic: ouch ... the redirect needs to contain the requesttoken. added it as a get param, but it requires encrypting the refreshtoken in the session. currently there is no way to officially get the CsrfTokenManager in an app ... meh ...

butonic:
had to rewrite the `HooksHander` to make things testable

lib/Controller/PasswordController.php
```diff
@@ -0,0 +1,167 @@
...
+	}
+	/**
+	 * Shows a change password template
+	 *
+	 * @param string $redirect_url
+	 * @return TemplateResponse|RedirectResponse|JSONResponse
+	 *
+	 * @throws \Exception
+	 */
+	public function show($redirect_url) {
```
DeepDiver1975: this method needs annotation NoCSRFRequired

appinfo/app.php
```diff
@@ -19,46 +19,51 @@
...
-	}
+	[$handler, 'saveOldPassword']
+);
+
+// FIXME postLogin is not only emitted in loginWithPassword, but also triggered
+// FIXME in loginWithToken, loginWithApache and loginUser, but does not make
+// FIXME sense in those cases
+
+// check password expiry after login
+OC::$server->getUserManager()->listen(
```
DeepDiver1975: emitter is not in the public api - no other way?

lib/Authentication/AccountModule.php
```diff
@@ -0,0 +1,85 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\PasswordPolicy\Authentication;
+
+use OC\Authentication\Exceptions\AccountCheckException;
```
DeepDiver1975: private API - we need to add this to OCP

lib/Controller/PasswordController.php
```diff
@@ -0,0 +1,167 @@
...
+	 * @param string $current_password
+	 * @param string $new_password
+	 * @param string $retyped_password
+	 * @param string $redirect_url
+	 * @return TemplateResponse|RedirectResponse
+	 *
+	 * @UseSession
+	 * @throws \Exception
+	 */
+	public function update($current_password, $new_password, $retyped_password, $redirect_url) {
```
DeepDiver1975: handle redirect url being null

butonic: cannot be null because the account module always adds a redirect url

appinfo/app.php
butonic: no, see FIXMEs above ...

lib/Controller/PasswordController.php
DeepDiver1975: hehe - again - I mean on the show method ;-)

lib/Authentication/AccountModule.php
butonic: hm yes. can you work on the core pr ... this needs a lot more tests.

lib/Controller/PasswordController.php
butonic: and show is called by the redirect request, which will contain a redirect url. if it doesn't someone is fiddling with the requests. he will see the force password change template ... but will not be redirected anywhere ... ok from my pov

DeepDiver1975: I can call show be entering the url in the browser - after changing the password I get a white page because the redirect holds no value.

lib/Authentication/AccountModule.php
butonic: new api in https://github.com/owncloud/core/pull/31915

lib/Controller/PasswordController.php
butonic: you are not supposed to enter that url in the browser

appinfo/app.php
```diff
@@ -19,46 +19,41 @@
...
-			$event->getArgument('password')
-		);
-	}
+	[$handler, 'saveOldPassword']
+);
+$eventDispatcher->addListener(
+	'user.afterlogin',
+	[$handler, 'checkPasswordExpired']
+);
+$eventDispatcher->addListener(
```
DeepDiver1975: merge with the listener above?

lib/Command/ExpirePassword.php
```diff
@@ -0,0 +1,104 @@
...
+			->addArgument(
+					'uid',
+					InputArgument::REQUIRED,
+					'the uid of the user in ownCloud'
+				     )
+			->addArgument(
+					'expiredate',
+					InputArgument::OPTIONAL,
+					'expire the password at the given date and time, eg. "2019-01-01 14:00:00 CET"',
+						'-1 days'		// base.php sets timezone to utc, so
```
DeepDiver1975: indent looks strange ....

lib/Command/ExpirePassword.php
```diff
@@ -0,0 +1,104 @@
...
+	 * @return int|null|void
+	 * @throws \OCP\PreConditionNotMetException
+	 */
+	protected function execute(InputInterface $input, OutputInterface $output) {
+		$uid = $input->getArgument('uid');
+
+		$exists = $this->userManager->userExists($uid);
+		if($exists === false) {
+			$output->writeln("<error>Unknown user $uid</error>");
+			return;
```
DeepDiver1975: use return codes

lib/Controller/PasswordController.php
```diff
@@ -0,0 +1,177 @@
...
+	 * @return TemplateResponse|JSONResponse
+	 *
+	 * @throws \Exception
+	 */
+	public function show($redirect_url) {
+		// gracefully handle redirected ocs api requests
+		if ($this->request->getHeader('OCS-APIREQUEST') === 'true') {
+			return new JSONResponse([], Http::STATUS_NO_CONTENT);
+		}
+		return $this->createPasswordTemplateResponse($redirect_url);
```
DeepDiver1975: set default redirect url if not given

lib/Controller/PasswordController.php
DeepDiver1975: the acceptance test will call the url directly ;-)
just add the default redirect url - we have this value
without setting the redirect url the redirect response is not valid

lib/Controller/PasswordController.php
```diff
@@ -0,0 +1,177 @@
...
+	 * @return TemplateResponse|RedirectResponse
+	 *
+	 * @UseSession
+	 * @throws \Exception
+	 */
+	public function update($current_password, $new_password, $retyped_password, $redirect_url) {
+		if ($new_password !== $retyped_password) {
+			return $this->createPasswordTemplateResponse(
+				$redirect_url,
+				$this->l10n->t('New passwords are not the same.')
```
DeepDiver1975: reads strange - we need string reviews @settermjd

lib/Controller/PasswordController.php
```diff
@@ -0,0 +1,177 @@
...
+			);
+		}
+
+		$user = $this->userSession->getUser();
+		if ($user === null) {
+			return $this->createPasswordTemplateResponse(
+				$redirect_url,
+				$this->l10n->t('No user in session.')
```
DeepDiver1975: user gives a :hankey: about the session - this message is to technical for the user

basically: if there is no user - no user is logged in -> throw unauthorized

lib/Controller/PasswordController.php
```diff
@@ -0,0 +1,177 @@
...
+			);
+		}
+
+		if(!$this->userManager->checkPassword($user->getUID(), $current_password)) {
+			return $this->createPasswordTemplateResponse(
+				$redirect_url,
+				$this->l10n->t('Wrong password.')
```
DeepDiver1975: which of the 3 the user entered?

lib/Controller/PasswordController.php
```diff
@@ -0,0 +1,177 @@
...
+			$changed = $user->setPassword($new_password);
+		} catch (\Exception $e) { // TODO introduce PolicyException that rules can throw
+			return $this->createPasswordTemplateResponse(
+				$redirect_url,
+				$e->getMessage()
+			);
+		}
+		if (!$changed) {
+			// threw exception, so the user can go to the admin
+			throw new \UnexpectedValueException($this->l10n->t('Password change failed.'));
```
DeepDiver1975: Changing the password failed

@settermjd for string review

lib/Controller/SettingsController.php
```diff
@@ -22,54 +22,54 @@
 namespace OCA\PasswordPolicy\Controller;
 
 use OCP\AppFramework\Controller;
+use OCP\AppFramework\Http\DataResponse;
```
DeepDiver1975: unused

lib/Controller/SettingsController.php
```diff
@@ -22,54 +22,54 @@
...
 
 class SettingsController extends Controller implements ISettings {
 
-	/** @var IRequest  */
-	protected $request;
 	/** @var IConfig  */
 	protected $config;
-	protected $appValues;
+
+	public const DEFAULTS = [
```
DeepDiver1975: visibility on const is php 7.1+ - intended? if so we need to adopt info.xml
I suggest to stay compatible with php 5.6

we need to add php-lint on php 5.6

lib/HooksHandler.php
```diff
@@ -21,118 +21,237 @@
...
+		$this->engine = $engine;
+		$this->hasher = $hasher;
+		$this->timeFactory = $timeFactory;
+		$this->l10n = $l10n;
+		$this->passwordExpiredRule = $passwordExpiredRule;
+		$this->oldPasswordMapper = $oldPasswordMapper;
+		$this->session = $session;
+	}
+
+	private function fixDI() {
```
DeepDiver1975: should not be necessary - instanciate this class using

```php

\OC::$server->query(HookManager::class);

File: lib/Rules/PasswordExpired.php [hunk 21]
```diff
@@ -0,0 +1,126 @@
...
+	protected $timeFactory;
+
+	/**
+	 * @param IL10N $l10n
+	 * @param ILogger $logger
+	 * @param OldPasswordMapper $mapper
+	 * @param IHasher $hasher
+	 * @param ITimeFactory $timeFactory
+	 */
+	public function __construct(IL10N $l10n, ILogger $logger, OldPasswordMapper $mapper, IHasher $hasher, ITimeFactory $timeFactory) {
```
DeepDiver1975: 80 chars

templates/password.php
```diff
@@ -0,0 +1,62 @@
...
+	/>
+	<label for="current_password" class="infield"><?php p($l->t('current password'));?></label>
+
+	<input type="password" id="new_password" name="new_password"
+		   value="" placeholder="<?php p($l->t('new password'));?>"
+	/>
+	<label for="new_password" class="infield"><?php p($l->t('new password'));?></label>
+
+	<input type="password" id="retyped_password" name="retyped_password"
+		   value="" placeholder="<?php p($l->t('retype new password'));?>"
```
DeepDiver1975: is *retyped* proper english? @settermjd

tests/Controller/PasswordControllerTest.php
```diff
@@ -0,0 +1,329 @@
...
+
+		$this->request = $this->createMock(IRequest::class);
+		$this->userSession = $this->createMock(IUserSession::class);
+		$this->userManager = $this->createMock(IUserManager::class);
+		$this->config = $this->createMock(IConfig::class);
+		$this->session = $this->createMock(ISession::class);
+		$this->urlGenerator = $this->createMock(IURLGenerator::class);
+		$this->l10n = $this->createMock(IL10N::class);
+		$this->l10n
+			->expects($this->any())
```
DeepDiver1975: expects any is default - you can just remove this line

tests/Db/OldPasswordMapperTest.php
```diff
@@ -20,10 +21,8 @@
  *
  */
 
-namespace OCA\PasswordPolicy\Tests\Db;
+namespace OCA\PasswordPolicy\Db;
```
DeepDiver1975: namespace does not match the folder - autoloader will not find this file.
it will not be autploaded - but still does not follow the rest of the code we have all over the place

tests/EngineTest.php
```diff
@@ -0,0 +1,132 @@
...
+ */
+
+use OCA\PasswordPolicy\Engine;
+use OCP\IDBConnection;
+use OCP\IL10N;
+use OCP\Security\IHasher;
+use OCP\Security\ISecureRandom;
+use Test\TestCase;
+
+class EngineTest extends TestCase {
```
DeepDiver1975: namespace missing

tests/HooksHandlerTest.php
```diff
@@ -0,0 +1,279 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\PasswordPolicy;
```
DeepDiver1975: test namespace - see above

tests/rules/PasswordExpiredTest.php
```diff
@@ -0,0 +1,183 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\PasswordPolicy\Rules;
```
DeepDiver1975: test namespace

tests/rules/PasswordExpiredTest.php
```diff
@@ -0,0 +1,183 @@
...
+
+use OCA\PasswordPolicy\Db\OldPasswordMapper;
+use OCA\PasswordPolicy\Db\OldPassword;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\IL10N;
+use OCP\ILogger;
+use OCP\Security\IHasher;
+
+class PasswordExpiredTest extends \PHPUnit_Framework_TestCase {
```
DeepDiver1975: use Tests\TestCase as base class please - please double check on other tests as well - THX

tests/rules/PasswordExpiredTest.php
butonic: Tests for a Class should reside in the same namespace, but finish with *Test, see https://github.com/owncloud/core/issues/30739#issue-303801425

tests/Db/OldPasswordMapperTest.php
butonic: autoloader never loads this file. phpunit does, Tests for a Class should reside in the same namespace, but finish with *Test, see https://github.com/owncloud/core/issues/30739#issue-303801425

appinfo/app.php
butonic: this way I can more test the two use cases with separate tests.

tests/rules/PasswordExpiredTest.php
DeepDiver1975: ^a proposed definition which is not yet approved or applied
in contrary we have naming conventions in place which we *try to* follow for years
We can change once decided - until then: please do as we do it now

lib/Controller/PasswordController.php
butonic: `Wrong old password.` sounds even worse. And if the new passwords mismatch it will be `New passwords are not the same.` or whatever we decide. If the new password fails the policies an explicit error will be reported ... so no change necessary IMO.

lib/Authentication/AccountModule.php
```diff
@@ -0,0 +1,75 @@
...
+			$redirectUrl = $this->request->getRequestUri();
+			$response = new RedirectResponse(
+				$this->urlGenerator->linkToRoute( 'password_policy.password.show', [
+						'redirect_url' => $redirectUrl
+					])
+			);
+			// full list of error codes https://msdn.microsoft.com/en-us/library/windows/desktop/ms681386(v=vs.85).aspx
+			// ldap relevant: https://github.com/rancher/rancher/issues/1941#issue-104315860
+			// we use 'user must reset password', not 'password expired' because the letter indicates password cannot be changed
+			throw new AccountCheckException($response, 'user must reset password', 0x773);
```
settermjd: How about lengthening the user message, slightly, to: "_The user must reset their password._"?

lib/Command/ExpirePassword.php
```diff
@@ -0,0 +1,108 @@
...
+	public function __construct(IConfig $config, IUserManager $userManager) {
+		parent::__construct();
+		$this->config = $config;
+		$this->userManager = $userManager;
+	}
+
+	protected function configure() {
+		$this
+			->setName('user:expire-password')
+			->setDescription('expire a password for a user')
```
settermjd: Would this not be better written as: "_Expire a user's password_"?

lib/Command/ExpirePassword.php
```diff
@@ -0,0 +1,108 @@
...
+			->addArgument(
+					'uid',
+					InputArgument::REQUIRED,
+					'the uid of the user in ownCloud'
```
settermjd: How about: "_The user's ownCloud uid._"?

lib/Command/ExpirePassword.php
```diff
@@ -0,0 +1,108 @@
...
+				     )
+			->addArgument(
+					'expiredate',
+					InputArgument::OPTIONAL,
+					'expire the password at the given date and time, eg. "2019-01-01 14:00:00 CET"',
```
settermjd: How about '_The date and time when the password expires. e.g., "2019-01-01 14:00:00 CET"._'

lib/Command/ExpirePassword.php
```diff
@@ -0,0 +1,108 @@
...
+	 * @param OutputInterface $output
+	 * @return int
+	 * @throws \OCP\PreConditionNotMetException
+	 */
+	protected function execute(InputInterface $input, OutputInterface $output) {
+		$uid = $input->getArgument('uid');
+
+		$exists = $this->userManager->userExists($uid);
+		if($exists === false) {
+			$output->writeln("<error>Unknown user $uid</error>");
```
settermjd: How about `<error>Unknown user: $uid</error>`

lib/Command/ExpirePassword.php
```diff
@@ -0,0 +1,108 @@
...
+		$this->config->setUserValue(
+			$uid,
+			'password_policy',
+			'forcePasswordChange',
+			$value
+		);
+
+		// show expire date if it was given
+		if ($input->hasArgument('expiredate')) {
+			$output->writeln("Password for $uid set to expire on ". $date->format('Y-m-d H:i:s T'));
```
settermjd: How about: "_The password for $uid is set to expire on_ "?

butonic:
Password reset Dialog
password reset

Mismatching passwords
password reset - mismatch

Policy error
password reset - policy

lib/Command/ExpirePassword.php
butonic: `expires. e.g., ` dot before e.g.?

lib/Authentication/AccountModule.php
butonic: I'll chinge it ... although this exception is only visible in the log ... and why would you not use `The user must reset *his* password.`

butonic: is their gender neutral?

lib/Controller/PasswordController.php
settermjd: How about: "_The two passwords are not the same._"

lib/Controller/PasswordController.php
```diff
@@ -0,0 +1,177 @@
...
+		if ($new_password !== $retyped_password) {
+			return $this->createPasswordTemplateResponse(
+				$redirect_url,
+				$this->l10n->t('New passwords are not the same.')
+			);
+		}
+
+		$user = $this->userSession->getUser();
+		if ($user === null) {
+			throw new \UnexpectedValueException('No user in session');
```
settermjd: How about: "_The user is not currently logged in._".

lib/Controller/PasswordController.php
settermjd: Instead of "_Wrong password._", how about "_Incorrect password supplied._". Even that suggestion doesn't quite feel right.

lib/Controller/PasswordController.php
```diff
@@ -0,0 +1,177 @@
...
+			$changed = $user->setPassword($new_password);
+		} catch (\Exception $e) { // TODO introduce PolicyException that rules can throw
+			return $this->createPasswordTemplateResponse(
+				$redirect_url,
+				$e->getMessage()
+			);
+		}
+		if (!$changed) {
+			// throw exception, so the user can go to the admin
+			throw new \UnexpectedValueException('Changing the password failed');
```
settermjd: How about "_Unable to update the user's password._"?

lib/HooksHandler.php
```diff
@@ -21,118 +21,242 @@
...
-			$l = \OC::$server->getL10NFactory()->get('password_policy');
+		if ($days !== null) {
+			$date = new \DateTime();
+			$date->setTime(0,0,0);
+			$date->add(new \DateInterval('P'.$days.'D'));
+
 			if ($params['expirationDate'] === null) {
 				$params['accepted'] = false;
-				$params['message'] = (string) $l->t('Expiration date is required');
+				$params['message'] = (string) $this->l10n->t('Expiration date is required');
```
settermjd: Just a small change "_An expiration date is required_".

lib/HooksHandler.php
```diff
@@ -21,118 +21,242 @@
...
-		$dbMapper = new OldPasswordMapper(\OC::$server->getDatabaseConnection());
-		$oldPassword = new OldPassword();
-		$oldPassword->setUid($user->getUID());
-		$oldPassword->setPassword(\OC::$server->getHasher()->hash($password));
-		$oldPassword->setChangeTime(\OC::$server->getTimeFactory()->getTime());
-		$dbMapper->insert($oldPassword);
+	private function getUser(GenericEvent $event) {
+		$user = $event->getArgument('user');
+		if(!$user instanceof IUser) {
+			throw new \UnexpectedValueException("user '$user' not an instance of IUser");
```
settermjd: "_The user, '$user', is not an instance of IUser._"

lib/HooksHandler.php
```diff
@@ -21,118 +21,242 @@
...
+				$this->forcePasswordChange();
+			}
+		}
+	}
+
+	public function checkForcePasswordChangeOnFirstLogin(GenericEvent $event) {
+		$this->fixDI();
+		$user = $event->getSubject();
+		if(!$user instanceof IUser) {
+			throw new \UnexpectedValueException("user '$user' not an instance of IUser");
```
settermjd: "_The user, '$user', is not an instance of IUser._"

lib/Rules/Length.php
```diff
@@ -23,10 +23,15 @@
...
+	 * @param $val
+	 * @throws PolicyException
+	 */
 	public function verify($password, $val) {
-		if (mb_strlen($password, 'UTF-8') < $val) {
-			throw new \Exception(
-				$this->l10n->t("Password is too short. Minimum %d characters are required.", [$val]));
+		if (\mb_strlen($password, 'UTF-8') < $val) {
+			throw new PolicyException(
+				$this->l10n->t('Password is too short. Minimum %d characters are required.', [$val]));
```
settermjd: "_The password is too short. At least %d characters are required._"

lib/Rules/Lowercase.php
```diff
@@ -23,19 +23,24 @@
...
+	 * @param $password
+	 * @param $val
+	 * @throws PolicyException
+	 */
 	public function verify($password, $val) {
 		if ($this->countLowercase($password) < $val) {
-			throw new \Exception(
-				$this->l10n->t("Password contains too few lowercase characters. Minimum %d lowercase characters are required.", [$val]));
+			throw new PolicyException(
+				$this->l10n->t('Password contains too few lowercase characters. Minimum %d lowercase characters are required.', [$val]));
```
settermjd: "_The password contains too few lowercase characters. At least %d lowercase characters are required._"

lib/Rules/Numbers.php
```diff
@@ -23,14 +23,19 @@
...
+	 * @param $password
+	 * @param $val
+	 * @throws PolicyException
+	 */
 	public function verify($password, $val) {
 		if ($this->countDigits($password) < $val) {
-			throw new \Exception(
-				$this->l10n->t("Password contains too few numbers. Minimum %d numbers are required.", [$val]));
+			throw new PolicyException(
+				$this->l10n->t('Password contains too few numbers. Minimum %d numbers are required.', [$val]));
```
settermjd: "_The password contains too few numbers. At least %d numbers are required._"

lib/Rules/PasswordExpired.php
```diff
@@ -0,0 +1,132 @@
...
+	 * @param string $password
+	 * @param int $days days until a password is considered expired
+	 * @param string $uid
+	 * @throws PolicyException
+	 */
+	public function verify($password, $days, $uid) {
+		$latestPassword = $this->mapper->getLatestPassword($uid);
+		if ($latestPassword === null) {
+			$this->logger->debug(
+				'Cannot determine age, no old password for uid {uid} found.',
```
settermjd: "_Cannot determine the password's age, as no existing password for uid {uid} was found._"

lib/Rules/PasswordExpired.php
```diff
@@ -0,0 +1,132 @@
...
+					'uid' => $latestPassword->getUid()
+				]
+			);
+			return;
+		}
+
+		$changed = $latestPassword->getChangeTime();
+		if ($changed === null) {
+			$this->logger->warning(
+				'Cannot determine age, old password {id} for uid {uid} has no change time.',
```
settermjd: "_Cannot determine the existing password's age, as the old password for uid {uid} has no change time._"

lib/Rules/PasswordExpired.php
```diff
@@ -0,0 +1,132 @@
...
+		$date = new \DateTime("@$changed");
+		$date->add(new \DateInterval("P{$days}D"));
+		$date->setTime(23, 59, 59); // include last day
+		$expiresAt = $date->getTimestamp();
+
+		$now = $this->timeFactory->getTime();
+
+		if ($expiresAt < $now) {
+			throw new PolicyException(
+				$this->l10n->t('Password older than %d days', [$days])
```
settermjd: "_The password is older than %d days._"

settermjd: "_The password is older than %d days._"

lib/Rules/PasswordExpired.php
```diff
@@ -0,0 +1,132 @@
...
+			);
+		}
+		$this->logger->debug(
+			'Password {id} for uid {uid} still valid for {seconds} seconds.',
```
settermjd: "_The password '{id}' for uid '{uid}' is still valid for {seconds} seconds._"

lib/Rules/PasswordHistory.php
```diff
@@ -60,8 +64,8 @@ public function verify($password, $val, $uid) {
 		);
 		foreach($oldPasswords as $oldPassword) {
 			if ($this->hasher->verify($password, $oldPassword->getPassword())) {
-				throw new \Exception(
-					$this->l10n->t("Password should be different than your last %d passwords", [$val]));
+				throw new PolicyException(
+					$this->l10n->t('Password should be different than your last %d passwords', [$val]));
```
settermjd: "_The password must be different to your previous %d passwords._"

lib/Rules/Special.php
```diff
@@ -27,28 +27,28 @@ class Special extends Base {
...
+			$allowedSpecialCharsAsArray = \str_split($allowedSpecialChars);
+			$s = \array_filter(\str_split($special), function($char) use ($allowedSpecialCharsAsArray){
+				return !\in_array($char, $allowedSpecialCharsAsArray, true);
 			});
-			if (count($s) > 0) {
-				throw new \Exception(
-					$this->l10n->t("Password contains invalid special characters. Only %s are allowed.", [$allowedSpecialChars]));
+			if (\count($s) > 0) {
+				throw new PolicyException(
+					$this->l10n->t('Password contains invalid special characters. Only %s are allowed.', [$allowedSpecialChars]));
```
settermjd: "_The password contains invalid special characters. Only %s are allowed._"

lib/Rules/Special.php
```diff
@@ -27,28 +27,28 @@ class Special extends Base {
...
 			}
 		}
-		if (strlen($special) < $val) {
-			throw new \Exception(
-				$this->l10n->t("Password contains too few special characters. Minimum %d special characters are required.", [$val]));
+		if (\strlen($special) < $val) {
+			throw new PolicyException(
+				$this->l10n->t('Password contains too few special characters. Minimum %d special characters are required.', [$val]));
```
settermjd: "_The password contains too few special characters. A minimum of %d special characters are required._"

lib/Rules/Uppercase.php
```diff
@@ -23,19 +23,24 @@
...
+	 * @param $password
+	 * @param $val
+	 * @throws PolicyException
+	 */
 	public function verify($password, $val) {
 		if ($this->countCapitals($password) < $val) {
-			throw new \Exception(
-				$this->l10n->t("Password contains too few uppercase characters. Minimum %d uppercase characters are required.", [$val]));
+			throw new PolicyException(
+				$this->l10n->t('Password contains too few uppercase characters. Minimum %d uppercase characters are required.', [$val]));
```
settermjd: "_The password contains too few uppercase characters. A minimum of %d uppercase characters are required._"

templates/password.php
settermjd: Yes. However, "confirm new password" seems to be the defacto standard.

DeepDiver1975:
THX @settermjd :+1:

PVince81:
Fixes https://github.com/owncloud/password_policy/issues/2

lib/Controller/PasswordController.php
butonic: This should never happen because the method is not annotated with `@PublicPage` so a user must be logged in. In turn the session must have a user set. I'll just delete this check because it only makes the code inspection happy, anyway.

lib/HooksHandler.php
```diff
@@ -21,118 +21,242 @@
...
 				$params['accepted'] = false;
-				$params['message'] = (string) $l->t('Expiration date is required');
+				$params['message'] = (string) $this->l10n->t('Expiration date is required');
 			}
 
 			// $date is the max expiration date
 			if ($date < $params['expirationDate']) {
 				$params['accepted'] = false;
-				$params['message'] = (string) $l->n('Expiration date cannot exceed %n day', 'Expiration date cannot exceed %n days', $days);
+				$params['message'] = (string) $this->l10n->n('Expiration date cannot exceed %n day', 'Expiration date cannot exceed %n days', $days);
```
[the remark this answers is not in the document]
butonic: When setting the expiration date of a link password the user can give a date. In the config the admin can limit the number of days a link password is valid by specifying e.g. 30 days. If we would give a specific date, say `The expiration date cannot exceed 2018.07.12.` that might be misunderstood as `All link passwords will expire on 2018.07.12.` So I think it makes more sense to tell the user that password protected links will automatically expire after e.g. 30 days. However, this check is now also reused fur user passwords, so the context is no longer limited to the former use case. I'll add `The`.

lib/HooksHandler.php
butonic: Technical exception. should never happen, but we check it to detect broken code flows. @settermjd why the two commes? You did not add them in the other text changes.

lib/Rules/PasswordExpired.php
```diff
@@ -0,0 +1,132 @@
...
+					'app' => __METHOD__,
+					'uid' => $uid
+				]
+			);
+			return;
+		}
+
+		if (!$this->hasher->verify($password, $latestPassword->getPassword())) {
+			$this->logger->warning(
+				'Cannot determine age, old password {id} for uid {uid} does not match entered password.',
```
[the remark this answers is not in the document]
butonic: I'll leave out the additional `'`s because id and uid are non empty strings and the additional `'` only make the log harder to read.

felixheidecke:
My attempt to review the UI :-)

lib/Controller/PasswordController.php
```diff
@@ -0,0 +1,174 @@
...
+	/**
+	 * Shows a change password template
+	 *
+	 * @param string $current_password
+	 * @param string $new_password
+	 * @param string $retyped_password
+	 * @param string $redirect_url
+	 * @return TemplateResponse|RedirectResponse
+	 *
+	 * @UseSession
```
DeepDiver1975: @butonic please add

NoAdminRequired annotation - THX

lib/Controller/PasswordController.php
```diff
@@ -0,0 +1,175 @@
...
+			$this->appName,
+			'password',
+			$params,
+			'guest'
+		);
+	}
+	/**
+	 * Shows a change password template
+	 *
+	 * @NoCSRFRequired
```
DeepDiver1975: NoAdminRequired missing here as well

butonic:
@felixheidecke I just rebuilt the functionality of AD ... I guess it is needed, in order to prevent someone else changing the password, because the session will always show you the password change dialog. In theory you could login and walk away. Then an attacker can change the password without having the credentials.

butonic:
all addressed. please review

DeepDiver1975:
can't we get the save button a bit prettier?

screenshot from 2018-07-03 15-06-58

DeepDiver1975:
screenshot from 2018-07-03 15-12-35

DeepDiver1975:
pushed the css change regarding the button

butonic:
@DeepDiver1975 um 'save' now is not centered.
@felixheidecke I agree that showing the rules makes sense. But we can do a complete mockup overhaul for that in a subsequent effort. Bluntly listing the rules that are in effect without making it look ugly might require some iterating. Happy to rewrite this in vue ... in a subsequent PR ;-)

felixheidecke:
bildschirmfoto 2018-07-04 um 10 15 44

lib/Controller/PasswordController.php
PVince81: added below

PVince81:
From what I see the texts got adjusted https://github.com/owncloud/password_policy/pull/15/files#diff-eccb6a6d787249c52045604418726ee9R34

Ready for merge ?

DeepDiver1975:
let me squash stuff a  it

Issue #20, opened 2018-07-06
Password error messages are inconsistent


phil-davis:
When the user has not entered enough [upper|lower|number|special] characters in a password, the error displayed sometimes says "At least..." and other times says "A minimum of..." - make it consistent. I like "At least..."

When the required number is 1, the error says something like:
```
At least 1 numbers are required.
```

Handle the singular case also, because actually that will happen a lot in real life.

phil-davis:
PR merged, works.

PR #21, opened 2018-07-06
Tidy password messages

phil-davis:
Fixes issue #20 

1) ``Engine.php`` was passing an empty array instead of empty string when the required special chars are not enabled - fix it.
2) When the user inputs the current password wrong, the message is "Incorrect password supplied". Be more explicit and say "Incorrect current password supplied".
3) Change "A minimum of..." to "At least..." in password error messages, making them all consistent.
4) Handle singular and plural for when only 1 or many of a character category are required.
5) When particular special characters are required, put those in the message so the user has some clue which special characters they have to use.

### DIFF
--- a/lib/Engine.php
+++ b/lib/Engine.php
@@ -130,7 +130,7 @@ class Engine {
 		}
 		if ($this->yes('spv_special_chars_checked')) {
 			$val = $this->configValues['spv_special_chars_value'];
-			$chars = [];
+			$chars = '';
 			if ($this->yes('spv_def_special_chars_checked')) {
 				$chars = $this->configValues['spv_def_special_chars_value'];
 			}
--- a/lib/Engine.php
+++ b/lib/Engine.php
@@ -109,27 +109,27 @@ class Engine {
 	 */
 	public function verifyPassword($password, $uid = null) {
 		if ($this->yes('spv_min_chars_checked')) {
-			$val = $this->configValues['spv_min_chars_value'];
+			$val = (int) $this->configValues['spv_min_chars_value'];
 			$r = new Length($this->l10n);
 			$r->verify($password, $val);
 		}
 		if ($this->yes('spv_lowercase_checked')) {
-			$val = $this->configValues['spv_lowercase_value'];
+			$val = (int) $this->configValues['spv_lowercase_value'];
 			$r = new Lowercase($this->l10n);
 			$r->verify($password, $val);
 		}
 		if ($this->yes('spv_uppercase_checked')) {
-			$val = $this->configValues['spv_uppercase_value'];
+			$val = (int) $this->configValues['spv_uppercase_value'];
 			$r = new Uppercase($this->l10n);
 			$r->verify($password, $val);
 		}
 		if ($this->yes('spv_numbers_checked')) {
-			$val = $this->configValues['spv_numbers_value'];
+			$val = (int) $this->configValues['spv_numbers_value'];
 			$r = new Numbers($this->l10n);
 			$r->verify($password, $val);
 		}
 		if ($this->yes('spv_special_chars_checked')) {
-			$val = $this->configValues['spv_special_chars_value'];
+			$val = (int) $this->configValues['spv_special_chars_value'];
 			$chars = '';
 			if ($this->yes('spv_def_special_chars_checked')) {
 				$chars = $this->configValues['spv_def_special_chars_value'];
@@ -138,7 +138,7 @@ class Engine {
 			$r->verify($password, $val, $chars);
 		}
 		if ($uid !== null && $this->yes('spv_password_history_checked')) {
-			$val = $this->configValues['spv_password_history_value'];
+			$val = (int) $this->configValues['spv_password_history_value'];
 			$dbMapper = new OldPasswordMapper($this->db);
 			$r = new PasswordHistory($this->l10n, $dbMapper, $this->hasher);
 			$r->verify($password, $val, $uid);
--- a/lib/Rules/Lowercase.php
+++ b/lib/Rules/Lowercase.php
@@ -24,14 +24,19 @@ namespace OCA\PasswordPolicy\Rules;
 class Lowercase extends Base {
 
 	/**
-	 * @param $password
-	 * @param $val
+	 * @param string $password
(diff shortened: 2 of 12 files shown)


phil-davis:
Happy for guidance here on how to "properly" do the plural translation stuff.

lib/Rules/Lowercase.php
```diff
@@ -30,8 +30,13 @@ class Lowercase extends Base {
 	 */
 	public function verify($password, $val) {
 		if ($this->countLowercase($password) < $val) {
-			throw new PolicyException(
-				$this->l10n->t('The password contains too few lowercase characters. At least %d lowercase characters are required.', [$val]));
+			if ($val > 1) {
```
DeepDiver1975: no - please use n() - this function was design to respect different language counting schemas.
Thanks for noticing - we always forget about this problem

lib/Rules/Numbers.php
```diff
@@ -30,8 +30,13 @@ class Numbers extends Base {
 	 */
 	public function verify($password, $val) {
 		if ($this->countDigits($password) < $val) {
-			throw new PolicyException(
-				$this->l10n->t('The password contains too few numbers. At least %d numbers are required.', [$val]));
+			if ($val > 1) {
```
DeepDiver1975: see above

lib/Rules/Special.php
```diff
@@ -42,8 +42,23 @@ public function verify($password, $val, $allowedSpecialChars) {
 			}
 		}
 		if (\strlen($special) < $val) {
-			throw new PolicyException(
-				$this->l10n->t('The password contains too few special characters. A minimum of %d special characters are required.', [$val]));
+			if (!empty($allowedSpecialChars)) {
+				if ($val > 1) {
```
DeepDiver1975: see above

lib/Rules/Uppercase.php
```diff
@@ -30,8 +30,13 @@ class Uppercase extends Base {
 	 */
 	public function verify($password, $val) {
 		if ($this->countCapitals($password) < $val) {
-			throw new PolicyException(
-				$this->l10n->t('The password contains too few uppercase characters. A minimum of %d uppercase characters are required.', [$val]));
+			if ($val > 1) {
```
DeepDiver1975: see above

DeepDiver1975:
> Happy for guidance here on how to "properly" do the plural translation stuff.

```php
$this->l10n->n('The expiration date cannot exceed %n day.', 'The expiration date cannot exceed %n days.', $days);

**Comment by DeepDiver1975** [comment]
we should have a very close look at all messages if we use plural forms properly - I doubt it ....

File: lib/Engine.php [hunk 5]
```diff
@@ -109,36 +109,36 @@ public function generatePassword() {
 	 */
 	public function verifyPassword($password, $uid = null) {
 		if ($this->yes('spv_min_chars_checked')) {
-			$val = $this->configValues['spv_min_chars_value'];
+			$val = (int) $this->configValues['spv_min_chars_value'];
```
phil-davis: This was nasty to find. The unit tests all pass, because they send actual numbers to ``verify($password, $val)``

https://github.com/owncloud/core/blob/master/lib/private/L10N/L10N.php#L103
Does an exact comparison of the count with ``1`` - so if you pass down a variable that is the string "1" then you end up being returned the plural translation anyway.
Perhaps that code in core could be more friendly and if some Wally has passed in "1" then send them back the singular translation.

phil-davis:
@DeepDiver1975 ready to review again.

lib/Rules/Lowercase.php
```diff
@@ -24,14 +24,19 @@
...
+	 * @param int $val
 	 * @throws PolicyException
 	 */
 	public function verify($password, $val) {
 		if ($this->countLowercase($password) < $val) {
 			throw new PolicyException(
-				$this->l10n->t('The password contains too few lowercase characters. At least %d lowercase characters are required.', [$val]));
+				$this->l10n->n(
+					'The password contains too few lowercase letters. At least 1 lowercase letter is required.',
+					'The password contains too few lowercase letters. At least %n lowercase letters are required ZZZ.',
```
DeepDiver1975: ZZZ :question:

lib/Rules/Lowercase.php
```diff
@@ -24,14 +24,19 @@
...
+	 * @param string $password
+	 * @param int $val
 	 * @throws PolicyException
 	 */
 	public function verify($password, $val) {
 		if ($this->countLowercase($password) < $val) {
 			throw new PolicyException(
-				$this->l10n->t('The password contains too few lowercase characters. At least %d lowercase characters are required.', [$val]));
+				$this->l10n->n(
+					'The password contains too few lowercase letters. At least 1 lowercase letter is required.',
```
DeepDiver1975: one instead of 1 :question:

lib/Rules/Lowercase.php
phil-davis: I left my crap in :(

lib/Rules/Lowercase.php
phil-davis: Updated

phil-davis:
@DeepDiver1975 and again, hopefully I got the crud out this time.

lib/Rules/Lowercase.php
```diff
@@ -24,14 +24,19 @@
...
 	 * @throws PolicyException
 	 */
 	public function verify($password, $val) {
 		if ($this->countLowercase($password) < $val) {
 			throw new PolicyException(
-				$this->l10n->t('The password contains too few lowercase characters. At least %d lowercase characters are required.', [$val]));
+				$this->l10n->n(
+					'The password contains too few lowercase letters. At least one lowercase letter is required.',
+					'The password contains too few lowercase letters. At least %n lowercase letters are required.',
+					(int) $val
```
karakayasemi: not important, but according to phpdoc val already int, looks like no need to cast again.

phil-davis: Well-spotted - that was from when I was debugging. Then at the end I made sure to always pass in ``int``

phil-davis: fixed

Issue #18, opened 2018-07-04
Notify when user password is about to expire or expired


PVince81:
### High level
- [x] User should be notified X days before the password expires to change the password (when remaining days is exactly X days): the notification mentions how many days (X). X defaults to 30 days
- [ ] ~~User should be notified every day before expiration after Y or less days are left (notifications can pile up), the notification mentions how many days left (it will count down every day). Y defaults to 7 days~~
- [x] User should be notified when the password has expired

### Tech detail
- [x] We need a cron job that reads the expiration values (wherever they are stored now) and sends a single notification.
- [x] Might need to write down a flag in the user preferences to track that a notification was already sent.
- [x] When changing the password, the flag must be reset to make sure notifications will be sent again.
- [x] Notification URL points to settings page to change user password
- [x] Notification must have an action set so that it also gets sent by email.
    - [ ] Action URL should point to the settings page for changing the password, if redirecting is possible.
    - [ ] Else: Have JS code in password policy app intercept the event for the button click and redirect to the settings page for changing password
- [ ] Skip any user from a backend that doesn't support the "set password" capability (expiration not support with LDAP for example)
- [ ] Trigger Symfony event for both events ?

phil-davis:
What does it mean in practice, that the user should be notified when the password has expired?
If it is an "internal" notification that comes, for example, on the webUI, then they will never see it. Because when they try to login, they will either be taken directly to a password change screen and have to change their password to login, or their account will have got locked out (if that is a possibility for users that wait ages after their password has expired)

PVince81:
We might need to add an action URL to the notification to make sure the notification system also sends it by email. The default option only sends by mail when an action is required). The URL should point at the page for changing the password if possible.

Now I'm not sure whether password expiry automatically kicks the users out of the system. If not, they might still see the notification.

jvillafanez:
> User should be notified Y days before the password expires to change the password

Is this hardcoded or configurable? I guess hardcoded for now, but configurable in the (near) future

> We need a cron job that reads the expiration values (wherever they are stored now) and sends a single notification.

Do we want a heavy long running job that runs once per day, or multiple smaller jobs that runs maybe once per hour?

> Notification must have an action URL so that it also gets sent by email. Action URL should point to the page for changing the password, if possible.

I guess we'll need to use a "go there" button or similar. However I'm not sure if that's going to work because the actions are supposed to be executed via ajax. This doesn't look good at first sight.

PVince81:
[...]

Issue #24, opened 2018-07-09
Password expiry for accounts that exist prior to the app being enabled


phil-davis:
Steps to reproduce

1. As admin, add a user account ``user123``
2. Login as ``user123`` and make sure things are "normal for you"
3. As admin, install and enable the ``password_policy`` app
4. Set a user password policy of ``xx`` days until user password expires
5. Wait ``xx`` days
6. Login as ``user123``

Expected behavior:
- the user is taken to the password change page

Actual behavior:
- the password continues to work

phil-davis:
To test this, I use my branch in PR #23 which shortens the password expiry to be in seconds, not days.
That works fine for me in ordinary testing.

phil-davis:
What is the requirement?

I suppose there is no record of data of last password change for existing accounts, so there is no way to accurately measure their password expiry for enforcing the first change.

Perhaps if there is no previous password change record for a user, then either:
1) trigger forced password change
or
2) when password_policy is enabled, remember somewhere the date it is enabled. Use that as the "default last password change" date. So existing users will get a password change triggered typically 90 days after the app was enabled.

Option (1) will trigger an immediate mass of forced password changes, resulting in a surge of help desk calls on the first day.

Option (2) will trigger the same surge towards the 90 day point.

PVince81:
@pmaier1 please clarify requirement.

I'd consider this a release blocker for the expiry as it would not work for existing users and will likely be reported as "broken"

pmaier1:
This boils down to: Password expiration policies only apply to users added after password_policy has been enabled? This would clearly be a release blocker. Expiration needs to apply to all (local) users no matter if they were created before or after activating the policy. Not sure how the point in time that starts the expiration period is defined. I would be ok with starting the period when the app is enabled, I think.

PVince81:
The point in time is based on the last record of whenever a user last changed their password.
Such record did not exist, so we need to decide.

As discussed in a chat, let's use the "date when the policy got activated" for every user where no record exists.

This is solution 2 as proposed by @phil-davis 

Since the app wasn't released yet, we can assume that no one has enabled the new policy yet in the real world so no migration is needed.

Estimate: 0.5md

pmaier1:
> As discussed in a chat, let's use the "date when the policy got activated" for every user where no record exists.

Sounds good from my POV. As this diverges from the original concept/implementation, @butonic please confirm.

butonic:
[...]

PR #27, opened 2018-07-09
send notification when passwords are about to expire

jvillafanez:
Fixes https://github.com/owncloud/password_policy/issues/18

### Open tasks
- [x] better SQL query
- [x] agree whether the JS hack is ok and/or whether we want to hide the "Change Password" button and exploit hack, see https://github.com/owncloud/password_policy/pull/27#issuecomment-403951044
- [x] delete related notifications after password was changed
- [x] solve confusing settings in page (PVince81 was confused)
- [x] adjust wording based on @pmaier1 feedback https://github.com/owncloud/password_policy/issues/18#issuecomment-403439985
- [x] unit tests
- [x] ~JS tests~

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -61,3 +61,16 @@ $eventDispatcher->addListener(
 	\OCP\IUser::class . '::firstLogin',
 	[$handler, 'checkForcePasswordChangeOnFirstLogin']
 );
+
+$app = new \OCA\PasswordPolicy\AppInfo\Application();
+$app->registerNotifier();
+
+// only load notification JS code in the logged in layout page (not public links not login page)
+$request = \OC::$server->getRequest();
+if (\OC::$server->getUserSession() !== null && \OC::$server->getUserSession()->getUser() !== null
+	&& substr($request->getScriptName(), 0 - strlen('/index.php')) === '/index.php'
+	&& substr($request->getPathInfo(), 0, strlen('/s/')) !== '/s/'
+	&& substr($request->getPathInfo(), 0, strlen('/login')) !== '/login') {
+
+	\OCP\Util::addScript('password_policy', 'notification');
+}
--- a/lib/Controller/SettingsController.php
+++ b/lib/Controller/SettingsController.php
@@ -26,6 +26,7 @@ use OCP\IConfig;
 use OCP\IRequest;
 use OCP\Settings\ISettings;
 use OCP\Template;
+use OCA\PasswordPolicy\UserNotificationConfigHandler;
 
 class SettingsController extends Controller implements ISettings {
 
@@ -49,11 +50,24 @@ class SettingsController extends Controller implements ISettings {
 		'spv_password_history_value' => 3,
 		'spv_user_password_expiration_checked' => false,
 		'spv_user_password_expiration_value' => 90,
+		'spv_user_password_expiration_notification_checked' => false,
+		'spv_user_password_expiration_notification_value' => UserNotificationConfigHandler::DEFAULT_EXPIRATION_FOR_NORMAL_NOTIFICATION,
 		'spv_user_password_force_change_on_first_login_checked' => false,
 		'spv_expiration_password_checked' => false,
 		'spv_expiration_password_value' => 7,
 		'spv_expiration_nopassword_checked' => false,
 		'spv_expiration_nopassword_value' => 7,
+	];
+
+	/**
+	 * functions to convert values between what is shown and what is stored
+	 * these functions must be defined in this class, they're per config key
+	 */
+	const CONVERSIONS = [
+		'spv_user_password_expiration_notification_value' => [
+			'in' => 'daysToSeconds',
+			'out' => 'secondsToDays',
+		],
 	];
 
 	public function __construct($appName,
@@ -71,6 +85,10 @@ class SettingsController extends Controller implements ISettings {
 			if ($this->request->getParam($key) !== null) {
 				if ($key !== 'spv_def_special_chars_value' && \substr($key, -6) === '_value') {
 					$value = \min(\max(0, (int)$this->request->getParam($key)), 255);
+					if (isset(self::CONVERSIONS[$key]['in'])) {
+						$convertFuncName = self::CONVERSIONS[$key]['in'];
+						$value = $this->$convertFuncName($value);
+					}
 					$this->config->setAppValue('password_policy', $key, $value);
 				} else {
 					$this->config->setAppValue('password_policy', $key, \strip_tags($this->request->getParam($key)));
@@ -92,9 +110,33 @@ class SettingsController extends Controller implements ISettings {
 	public function getPanel() {
 		$template = new Template('password_policy', 'admin');
 		foreach(self::DEFAULTS as $key => $default) {
-			$template->assign($key, $this->config->getAppValue('password_policy', $key, $default));
+			$value = $this->config->getAppValue('password_policy', $key, $default);
+			if (isset(self::CONVERSIONS[$key]['out'])) {
+				$convertFuncName = self::CONVERSIONS[$key]['out'];
+				$value = $this->$convertFuncName($value);
+			}
+			$template->assign($key, $value);
 		}
 		return $template;
 	}
 
+	/**
+	 * Convert the days to seconds
+	 * @param int $days
+	 * @return int the number of seconds
+	 */
+	private function daysToSeconds($days) {
+		return $days * 24 * 60 * 60;
+	}
+
+	/**
+	 * Convert seconds to days. The value will always be rounded up,
+	 * so 1 second will be converted to 1 day
+	 * @param int $seconds the number of seconds to be converted
+	 * @return int the number of days in those seconds, rounded up
+	 */
+	private function secondsToDays($seconds) {
+		$floatDays = $seconds / (24 * 60 * 60);
+		return \intval(\ceil($floatDays));
+	}
 }
--- a/lib/Db/OldPasswordMapper.php
+++ b/lib/Db/OldPasswordMapper.php
@@ -87,7 +87,7 @@ class OldPasswordMapper extends Mapper {
 			$info = \json_encode($stmt->erroInfo());
 			$message = "Cannot get the passwords that are about to expire. Error: {$info}";
 			\OCP\Util::writeLog('password_policy', $message, \OCP\Util::ERROR);
-			return false;
+			return;
 		}
 
 		while ($row = $stmt->fetch()) {
--- a/lib/Db/OldPasswordMapper.php
+++ b/lib/Db/OldPasswordMapper.php
@@ -63,4 +63,39 @@ class OldPasswordMapper extends Mapper {
 		}
 		return $passwords[0];
 	}
+
+	/**
+	 * Get the passwords that are about to expire or already expired.
+	 * Last passwords which have been changed before the timestamp are the ones
+	 * selectable. Previous stored passwords won't be included
+	 * In addition, passwords from multiple users are expected
+	 * @param int $maxTimestamp timestamp marker, last passwords changed before
+	 * the timestamp will be selected
+	 * @return OldPassword[] the selected passwords
+	 */
+	public function getPasswordsAboutToExpire($maxTimestamp) {
+		$oldPasswords = [];
+
+		$query = "SELECT `f`.`id`, `f`.`uid`, `f`.`password`, `f`.`change_time` FROM (";
+		$query .= "SELECT `uid`, max(`change_time`) AS `maxtime` FROM `*PREFIX*user_password_history` GROUP BY `uid`";
+		$query .= ") AS `x` INNER JOIN `*PREFIX*user_password_history` AS `f` ON `f`.`uid` = `x`.`uid` AND `f`.`change_time` = `x`.`maxtime`";
+		$query .= " WHERE `f`.`change_time` < ?";
+
+		$stmt = $this->db->prepare($query);
+		$stmt->bindValue(1, $maxTimestamp);
+		$result = $stmt->execute();
+
+		if ($result === false) {
+			$info = \json_encode($stmt->erroInfo());
+			$message = "Cannot get the passwords that are about to expire. Error: {$info}";
+			\OCP\Util::writeLog('password_policy', $message, \OCP\Util::ERROR);
+			return false;
+		}
+
+		while ($row = $stmt->fetch()) {
+			$oldPasswords[] = OldPassword::fromRow($row);
+		}
+		$stmt->closeCursor();
+		return $oldPasswords;
+	}
 }
--- a/lib/Db/OldPasswordMapper.php
+++ b/lib/Db/OldPasswordMapper.php
@@ -71,11 +71,9 @@ class OldPasswordMapper extends Mapper {
 	 * In addition, passwords from multiple users are expected
 	 * @param int $maxTimestamp timestamp marker, last passwords changed before
 	 * the timestamp will be selected
-	 * @return OldPassword[] the selected passwords
+	 * @return Generator to traverse the selected passwords
 	 */
 	public function getPasswordsAboutToExpire($maxTimestamp) {
-		$oldPasswords = [];
-
 		$query = "SELECT `f`.`id`, `f`.`uid`, `f`.`password`, `f`.`change_time` FROM (";
 		$query .= "SELECT `uid`, max(`change_time`) AS `maxtime` FROM `*PREFIX*user_password_history` GROUP BY `uid`";
 		$query .= ") AS `x` INNER JOIN `*PREFIX*user_password_history` AS `f` ON `f`.`uid` = `x`.`uid` AND `f`.`change_time` = `x`.`maxtime`";
@@ -93,9 +91,8 @@ class OldPasswordMapper extends Mapper {
 		}
 
 		while ($row = $stmt->fetch()) {
-			$oldPasswords[] = OldPassword::fromRow($row);
+			yield OldPassword::fromRow($row);
 		}
 		$stmt->closeCursor();
-		return $oldPasswords;
 	}
 }
--- a/lib/Jobs/PasswordExpirationNotifierJob.php
+++ b/lib/Jobs/PasswordExpirationNotifierJob.php
@@ -107,6 +107,15 @@ class PasswordExpirationNotifierJob extends TimedJob {
 		}
 	}
 
+	/**
+	 * Send an "about to expire" notification using the password information
+	 * in $passInfo. The password should expire after $expirationTime (90 days
+	 * by default). This information will also be used in the notification
+	 * @param OldPassword $passInfo the password information used to send the
+	 * notification
+	 * @param int $expirationTime the time to expire the password in seconds
+	 * (for example, 90 days - in seconds)
+	 */
 	private function sendAboutToExpireNotification(OldPassword $passInfo, $expirationTime) {
 		if ($this->unConfigHandler->isSentAboutToExpireNotification($passInfo)) {
 			return;  // notification already sent
@@ -134,6 +143,15 @@ class PasswordExpirationNotifierJob extends TimedJob {
 		$this->unConfigHandler->markAboutToExpireNotificationSentFor($passInfo);
 	}
 
+	/**
+	 * Send an "expired" notification using the password information
+	 * in $passInfo. The password should expire after $expirationTime (90 days
+	 * by default). This information will also be used in the notification
+	 * @param OldPassword $passInfo the password information used to send the
+	 * notification
+	 * @param int $expirationTime the time to expire the password in seconds
+	 * (for example, 90 days - in seconds)
+	 */
 	private function sendPassExpiredNotification(OldPassword $passInfo, $expirationTime) {
 		if ($this->unConfigHandler->isSentExpiredNotification($passInfo)) {
 			return;  // notification already sent
--- a/lib/Jobs/PasswordExpirationNotifierJob.php
+++ b/lib/Jobs/PasswordExpirationNotifierJob.php
@@ -1,0 +1,170 @@
+<?php
+/**
+ *
+ * @author Juan Pablo Villafáñez <jvillafanez@solidgear.es>
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license GPL-2.0
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\PasswordPolicy\Jobs;
+
+use OC\BackgroundJob\TimedJob;
+use OCP\Notification\IManager;
+use OCP\IURLGenerator;
+use OCP\ILogger;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCA\PasswordPolicy\Db\OldPasswordMapper;
+use OCA\PasswordPolicy\Db\OldPassword;
+use OCA\PasswordPolicy\UserNotificationConfigHandler;
+
+class PasswordExpirationNotifierJob extends TimedJob {
+	/** @var OldPasswordMapper */
+	private $mapper;
+
+	/** @var $manager */
+	private $manager;
+
+	/** @var UserNotificationConfigHandler */
+	private $unConfigHandler;
+
+	/** @var ITimeFactory */
+	private $timeFactory;
+
+	/** @var IURLGenerator */
+	private $urlGenerator;
+
+	/** @var ILogger */
+	private $logger;
+
+	public function __construct(
+		OldPasswordMapper $mapper,
+		IManager $manager,
+		UserNotificationConfigHandler $unConfigHandler,
+		ITimeFactory $timeFactory,
+		IURLGenerator $urlGenerator,
+		ILogger $logger
+	) {
+		$this->mapper = $mapper;
+		$this->manager = $manager;
+		$this->unConfigHandler = $unConfigHandler;
+		$this->timeFactory = $timeFactory;
+		$this->urlGenerator = $urlGenerator;
+		$this->logger = $logger;
+
+		$this->setInterval(12 * 60 * 60);
+	}
+
+	protected function run($arguments) {
+		$expirationTime = $this->unConfigHandler->getExpirationTime();
+		if ($expirationTime === null) {
+			return;  // expiration not configured
+		}
+
+		$expirationTimeNotification = $this->unConfigHandler->getExpirationTimeForNormalNotification();
+		if ($expirationTimeNotification === null) {
+			$expirationTimeNotification = 0;
+		}
+
+		// ensure ranges are correct
+		if ($expirationTime <= $expirationTimeNotification) {
+			$message = "wrong expiration range: normal ($expirationTimeNotification) < expired ($expirationTime)";
+			$this->logger->warning($message, ['app' => 'password_policy']);
+			return;
+		}
+
+		$notifyAfter = $expirationTime - $expirationTimeNotification;
+
+		$currentTime = $this->timeFactory->getTime();
+
+		$maxTimestamp = $currentTime - $notifyAfter;
+		// passwords changed before $maxTimestamp are expired or about to be expired
+		// according to the expiration time range
+
+		$oldPasswordsAboutToExpire = $this->mapper->getPasswordsAboutToExpire($maxTimestamp);
+		foreach ($oldPasswordsAboutToExpire as $passInfo) {
+			$elapsedTime = $currentTime - $passInfo->getChangeTime();
+			if ($elapsedTime >= $expirationTime) {
+				$this->logger->debug("password timestamp for {$passInfo->getUid()}: {$passInfo->getChangeTime()}; elapsed time: {$elapsedTime} -> EXPIRED", ['app' => 'password_policy']);
+				$this->sendPassExpiredNotification($passInfo, $expirationTime);
+			} elseif ($elapsedTime >= $notifyAfter) {
+				$this->logger->debug("password timestamp for {$passInfo->getUid()}: {$passInfo->getChangeTime()}; elapsed time: {$elapsedTime} -> NOTIFY", ['app' => 'password_policy']);
+				$this->sendAboutToExpireNotification($passInfo, $expirationTime);
+			}
+		}
+	}
+
+	private function sendAboutToExpireNotification(OldPassword $passInfo, $expirationTime) {
+		if ($this->unConfigHandler->isSentAboutToExpireNotification($passInfo)) {
+			return;  // notification already sent
+		}
+
+		$notificationTimestamp = $this->timeFactory->getTime();
+
+		// we'll use the id of the passInfo as object id and marker
+		$notification = $this->manager->createNotification();
+		$notification->setApp('password_policy')
+			->setUser($passInfo->getUid())
+			->setDateTime(new \DateTime("@{$notificationTimestamp}"))
+			->setObject('about_to_expire', $passInfo->getId())
+			->setSubject('about_to_expire', [$passInfo->getChangeTime(), $expirationTime])
+			->setMessage('about_to_expire', [$passInfo->getChangeTime(), $expirationTime])
+			->setLink($this->getNotificationLink());
+
+		$linkAction = $notification->createAction();
+		$linkAction->setLabel('Change password')
+			->setLink($this->getNotificationLink(), 'GET');
+		$notification->addAction($linkAction);
+
+		$this->manager->notify($notification);
+
+		$this->unConfigHandler->markAboutToExpireNotificationSentFor($passInfo);
+	}
+
+	private function sendPassExpiredNotification(OldPassword $passInfo, $expirationTime) {
+		if ($this->unConfigHandler->isSentExpiredNotification($passInfo)) {
+			return;  // notification already sent
+		}
+
+		$notificationTimestamp = $this->timeFactory->getTime();
+
+		// we'll use the id of the passInfo as object id and marker
+		$notification = $this->manager->createNotification();
+		$notification->setApp('password_policy')
+			->setUser($passInfo->getUid())
+			->setDateTime(new \DateTime("@{$notificationTimestamp}"))
+			->setObject('expired', $passInfo->getId())
+			->setSubject('expired', [$passInfo->getChangeTime(), $expirationTime])
+			->setMessage('expired', [$passInfo->getChangeTime(), $expirationTime])
+			->setLink($this->getNotificationLink());
+
+		$linkAction = $notification->createAction();
+		$linkAction->setLabel('Change password')
+			->setLink($this->getNotificationLink(), 'GET');
+		$notification->addAction($linkAction);
+
+		$this->manager->notify($notification);
+
+		$this->unConfigHandler->markExpiredNotificationSentFor($passInfo);
+	}
+
+	private function getNotificationLink() {
+		return $this->urlGenerator->linkToRouteAbsolute(
+			'settings.SettingsPage.getPersonal',
+			['sectionid' => 'general']
+		);
+	}
+}
--- a/lib/Notifier.php
+++ b/lib/Notifier.php
@@ -24,7 +24,7 @@ use OCP\Notification\INotifier;
 use OCP\Notification\IManager as INotificationManager;
 use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\L10N\IFactory;
-use OC\L10N\L10N;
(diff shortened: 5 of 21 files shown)


lib/Db/OldPasswordMapper.php
```diff
@@ -63,4 +63,30 @@ public function getLatestPassword($uid) {
 		}
 		return $passwords[0];
 	}
+
+	public function getPasswordsAboutToExpire($maxTimestamp) {
+		$oldPasswords = [];
+
+		$query = "select `f`.`*` from (";
```
PVince81: I'd prefer using a simpler query like `SELECT uid, MAX(change_time) as `maxtime` FROM`*PREFIX*user_password_history` GROUP BY uid HAVING MAX(change_time) < ? `

It was pointed to me that the reason for this is because the `OldPassword` object requires the original fields to be present in the select, so this JOIN is here just to achieve this.

I suggested skipping the OldPassword object here and just returning a list of users. This wouldn't match well with the "ObjectMapper" concept. So maybe we should move this method elsewhere ?

Our problem currently is to extend this query to also cover cases where no change_time is set for https://github.com/owncloud/password_policy/issues/24

PVince81: @jvillafanez how about adding a new method on `OldPassword` that lets you create an instance with a given id ? Then here you could simplify the query as discussed instead of having to contraint it to `OldPassword::fromRow()`

PVince81: it should be possible to include all information in the original query somehow

let's continue for now and if needed we can ask @butonic to help us optimize this

unit tests will prove that it still works regardless

PVince81:
Query doesn't work on MariaDB:
```
MariaDB [owncloud]> select `f`.`*` from (select `uid`, max(`change_time`) as `maxtime` from `oc_user_password_history` group by `uid`) as `x` inner join `oc_user_password_history` as `f` on `f`.`uid` = `x`.`uid` and `f`.`change_time` = `x`.`maxtime` where `f`.`change_time` < 90;
ERROR 1054 (42S22): Unknown column 'f.*' in 'field list'
```

PVince81:
- [ ] discard notifications after password change, if possible. maybe need to store notification id in the oc_preferences ?

PVince81:
It seems the notifications app has a bug: when only a single action is specified, it doesn't display any action button... doesn't even seem to return the actions structure through the OCS call for some reason.
The good news is that the email is sent anyway.

So while we can rely on this bug for now, we do need to think about what happens if the button ever becomes visible. I'll still investigate the JS approach.

PVince81:
Ok, I debugged a bit and found out why: we need to implement a `Notifier` to be able to translate the texts.

Currently the action button does not appear because we don't have "parsed actions". This also means that we could deliberately not return a parsed action so no button will be displayed.

The email sending code from the notifications app relies on regular action, not parsed action.
Raised here: https://github.com/owncloud/notifications/issues/213

PVince81:
@jvillafanez look at this beautiful hack: https://github.com/owncloud/password_policy/pull/27/commits/be90884d11094490c6bcb1c73b1aba232a6ce287

I've enabled the parsed actions so the button becomes visible. If we want to, we could decide to still disable it so the button doesn't appear and exploit bug https://github.com/owncloud/notifications/issues/213

Now with the button visible, the notifications app only sends out the "OCA.Notifications.Action" jquery event if the ajax call was successful. So here I've set the link action to points at the settings page with GET, which will succeed anyway with 200 OK. After success, we catch the event in the password policy's "js/notification.js" and actually redirect to the link.

Nice trick but I hope we can make redirects work officially with https://github.com/owncloud/notifications/issues/212

One missing part is that clicking on "Change Password" doesn't automatically delete the notification. This could be acceptable as we should delete the notifications after the password was changed.

appinfo/app.php
```diff
@@ -57,3 +57,15 @@
...
 );
+
+$app = new \OCA\PasswordPolicy\AppInfo\Application();
+$app->registerNotifier();
+
+$request = \OC::$server->getRequest();
+if (\OC::$server->getUserSession() !== null && \OC::$server->getUserSession()->getUser() !== null
+	&& substr($request->getScriptName(), 0 - strlen('/index.php')) === '/index.php'
+	&& substr($request->getPathInfo(), 0, strlen('/s/')) !== '/s/'
+	&& substr($request->getPathInfo(), 0, strlen('/login')) !== '/login') {
```
phil-davis: Is the reason for this ``if`` condition obvious? I could use a comment, or some way to understand what is the script+path pattern aimed for and why.

lib/Jobs/PasswordExpirationNotifierJob.php
```diff
@@ -0,0 +1,169 @@
...
+		$this->setInterval(12 * 60 * 60);
+	}
+
+	protected function run($arguments) {
+		$expirationTime = $this->unConfigHandler->getExpirationTime();
+		if ($expirationTime === null) {
+			return;  // expiration not configured
+		}
+
+		$expirationTime = $expirationTime * 24 * 60 * 60;  // convert days to seconds
```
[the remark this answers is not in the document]
phil-davis: There is a ``daysToSeconds`` function in another file above. I wonder how we can contain these kind of unit-of-measure conversions to some always-accessible functions, rather than having mentions of ``24`` and ``60`` around the code?

lib/UserNotificationConfigHandler.php
```diff
@@ -0,0 +1,118 @@
...
+ *
+ */
+
+namespace OCA\PasswordPolicy;
+
+use OCP\IConfig;
+use OCA\PasswordPolicy\Db\OldPassword;
+
+class UserNotificationConfigHandler {
+	const DEFAULT_EXPIRATION_FOR_NORMAL_NOTIFICATION = 30 * 24 * 60 * 60;  // 30 days
```
phil-davis: Another place where having the "magic" ``24 * 60 * 60`` somewhere common will help for the future.

lib/UserNotificationConfigHandler.php
```diff
@@ -0,0 +1,118 @@
...
+	public function markExpiredNotificationSentFor(OldPassword $passInfo) {
+		$this->config->setUserValue($passInfo->getUid(), 'password_policy', 'expiredSent', $passInfo->getId());
+	}
+
+	public function isSentAboutToExpireNotification(OldPassword $passInfo) {
+		$storedId = $this->config->getUserValue($passInfo->getUid(), 'password_policy', 'aboutToExpireSent', null);
+		if ($storedId === null) {
+			return false;  // notification not sent
+		} elseif (\intval($storedId) !== $passInfo->getId()) {
+			// if the pasword id doesn't match the one stored, the notification hasn't been sent
```
phil-davis: s/pasword/password

lib/UserNotificationConfigHandler.php
```diff
@@ -0,0 +1,118 @@
...
+		}
+		return true;
+	}
+
+	public function isSentExpiredNotification(OldPassword $passInfo) {
+		$storedId = $this->config->getUserValue($passInfo->getUid(), 'password_policy', 'expiredSent', null);
+		if ($storedId === null) {
+			return false;  // notification not sent
+		} elseif (\intval($storedId) !== $passInfo->getId()) {
+			// if the pasword id doesn't match the one stored, the notification hasn't been sent
```
phil-davis: s/pasword/password

tests/NotifierTest.php
```diff
@@ -0,0 +1,155 @@
...
+		$expireIn = 10 * 24 * 60 * 60;  // 10 days;
+		$notification = $this->createMock(INotification::class);
+		$notification->method('getApp')->willReturn('password_policy');
+		$notification->method('getObjectType')->willReturn('about_to_expire');
+		$notification->method('getSubjectParameters')->willReturn([$initialTime, $expireIn]);
+		// first parameter is the password change time and the second is the
+		// expiration time starting from the password change time
+		$notification->method('getMessageParameters')->willReturn([$initialTime, $expireIn]);
+
+		$this->timeFactory->method('getTime')->willReturn($initialTime + (7 * 24 * 60 * 60));
```
phil-davis: Another days to seconds conversion.

tests/NotifierTest.php
```diff
@@ -0,0 +1,155 @@
...
+		$expireIn = 10 * 24 * 60 * 60;  // 10 days;
+		$notification = $this->createMock(INotification::class);
+		$notification->method('getApp')->willReturn('password_policy');
+		$notification->method('getObjectType')->willReturn('about_to_expire');
+		$notification->method('getSubjectParameters')->willReturn([$initialTime, $expireIn]);
+		// first parameter is the password change time and the second is the
+		// expiration time starting from the password change time
+		$notification->method('getMessageParameters')->willReturn([$initialTime, $expireIn]);
+
+		$this->timeFactory->method('getTime')->willReturn($initialTime + (12 * 24 * 60 * 60));
```
phil-davis: Another days to seconds conversion.

tests/NotifierTest.php
```diff
@@ -0,0 +1,155 @@
...
+
+		$notification->expects($this->once())
+			->method('setParsedSubject')
+			->with('Your password is about to expire!');
+
+		$notification->expects($this->once())
+			->method('setParsedMessage')
+			->with('Your password has already expired 2 days ago');
```
phil-davis: Hmmm - wherever in the real code this message is generated, the text could be more simple and straightforward:
``Your password expired 2 days ago``

lib/Notifier.php
```diff
@@ -0,0 +1,132 @@
...
+	private function formatExpired(INotification $notification, L10N $l) {
+		$params = $notification->getSubjectParameters();
+		$notification->setParsedSubject(
+			(string) $l->t('Your password has expired', $params)
+		);
+
+		$messageParams = $notification->getMessageParameters();
+
+		$notification->setParsedMessage(
+			(string) $l->t('You have to change your password before you can access again', $messageParams)
```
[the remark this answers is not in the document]
PVince81: @pmaier1

appinfo/app.php
PVince81: copy-pasted from notifications app: https://github.com/owncloud/notifications/blob/master/appinfo/app.php#L29

I'll add a comment

PVince81:
Typos and texts adjusted

@jvillafanez please look into:
- [ ] mix up of "days to seconds" in various places, see comments

lib/Jobs/PasswordExpirationNotifierJob.php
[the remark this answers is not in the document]
jvillafanez: I expect a follow up PR to store the expirationTime in seconds soon. Once it's done we can remove this line and move on. I'd prefer to keep this for now.
It this were a long time solution, I'd definitely agree with you.

lib/Jobs/PasswordExpirationNotifierJob.php
```diff
@@ -0,0 +1,169 @@
...
+		$expirationTimeNotification = $this->unConfigHandler->getExpirationTimeForNormalNotification();
+		if ($expirationTimeNotification === null) {
+			$expirationTimeNotification = 0;
+		}
+
+		// ensure ranges are correct
+		if ($expirationTime <= $expirationTimeNotification) {
```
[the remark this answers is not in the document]
jvillafanez: I'll include a PHPDoc in the functions to clarify

lib/Jobs/PasswordExpirationNotifierJob.php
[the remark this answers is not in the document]
jvillafanez: In fact, it makes more sense to move this conversion to the `getExpirationTime` function

lib/Notifier.php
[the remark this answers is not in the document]
jvillafanez: I think the product name is already shown in the mail template... Note that the message will be shown in the notification app too (already inside ownCloud)

lib/UserNotificationConfigHandler.php
jvillafanez: I think this is fine because you see clearly the conversion. The other option is to write directly the number of seconds, but it will be a random number for a lot of people and we'll need to trust the comment.

jvillafanez:
I think the "daysToSeconds" problem should be fine now.
* `DEFAULT_EXPIRATION_FOR_NORMAL_NOTIFICATION` constant is in seconds, but shown like that to make it easy to know that those are 30 days
* The conversion that was in the cron job has been moved to the `UserNotificationConfigHandler` to return seconds instead of days, so it's more consistent with the rest of the functions. Note that the conversion is currently required as the value is stored in days, but it's expected to store in seconds at some point.
* The ones for the test should be fine. The tested functions work with seconds, but messages refer to days.

jvillafanez:
> discard notifications after password change, if possible. maybe need to store notification id in the oc_preferences ?

The notification is discarded after clicking the action. The problem probably is that the redirection prevent further code execution.
Code flow is:
1. Trigger event
2. Respond to event (by all the event listeners)
3. Delete notification

If a redirection happens in 2, the deletion won't happen (or so it seems). Delaying the redirection should do the trick.

lib/Jobs/PasswordExpirationNotifierJob.php
```diff
@@ -0,0 +1,168 @@
...
+			->setLink($this->getNotificationLink(), 'GET');
+		$notification->addAction($linkAction);
+
+		$this->manager->notify($notification);
+
+		$this->unConfigHandler->markExpiredNotificationSentFor($passInfo);
+	}
+
+	private function getNotificationLink() {
+		return $this->urlGenerator->linkToRouteAbsolute(
```
jvillafanez: I guess this is fine for an "about to expire" notification, but if the password is expired maybe we should send a link to the password policy app instead, as I expect this link won't be accesible for the user.

jvillafanez:
> delete related notifications after password was changed

This should be fixed with https://github.com/owncloud/password_policy/pull/27/commits/b8c1a006418cff08c1f5fc0fe686175243893f37 although the solution is quite dirty. We should schedule some time to find a proper fix.

lib/UserNotificationConfigHandler.php
phil-davis: No problem - I agree that writing something like ``30*86400`` is not a good thing!

jvillafanez:
This should be ready unless someone says otherwise.

We'll need to schedule some time to refactor the part of the notifications because of the unittests, but I'll need a bit of time to plan how to do it properly. In any case, the code should be working, so we can do it after merging.

phil-davis:
I will give it a manual test in the morning and see what I can break.

lib/Jobs/PasswordExpirationNotifierJob.php
```diff
@@ -0,0 +1,170 @@
...
+		$linkAction->setLabel('Change password')
+			->setLink($this->getNotificationLink(), 'GET');
+		$notification->addAction($linkAction);
+
+		$this->manager->notify($notification);
+
+		$this->unConfigHandler->markAboutToExpireNotificationSentFor($passInfo);
+	}
+
+	private function sendPassExpiredNotification(OldPassword $passInfo, $expirationTime) {
```
PVince81: PHPDoc ?

lib/Jobs/PasswordExpirationNotifierJob.php
PVince81: if the password is expired the session might still be there, so the user will get to the page

now if the user got kicked out, opening the URL will display the login page + "change your password" full screen page. And then the user lands on the settings page. This is ok for now as discussed with @pmaier1 before.

lib/Notifier.php
```diff
@@ -0,0 +1,132 @@
...
+			case 'about_to_expire':
+				return $this->formatAboutToExpire($notification, $l);
+			case 'expired':
+				return $this->formatExpired($notification, $l);
+			default:
+				throw new \InvalidArgumentException();
+		}
+	}
+
+	private function formatAboutToExpire(INotification $notification, L10N $l) {
```
PVince81: PHPDoc. Should it be IL10N ?

lib/UserNotificationConfigHandler.php
```diff
@@ -0,0 +1,183 @@
...
+			return null;  // passwords don't expire or have weird value
+		}
+		// the expiration time is currently stored in days, so we need to convert
+		// it to seconds.
+		return \intval($expirationTime) * 24 * 60 * 60;
+	}
+
+	/**
+	 * Return the number of seconds until a user should receive a notification
+	 * that his password is about to expire. This _should_ be less than the value
```
PVince81: s/his/their/

templates/admin.php
```diff
@@ -90,6 +90,13 @@
 					<span><?php p($l->t('days until user password expires'));?></span>
 				</label>
 			</li>
+			<li>
+				<label>
+					<input type="checkbox" name="spv_user_password_expiration_notification_checked" <?php if ($_['spv_user_password_expiration_notification_checked']): ?> checked="checked"<?php endif; ?>/>
+					<input type="number" name="spv_user_password_expiration_notification_value"  min="0" value="<?php p($_['spv_user_password_expiration_notification_value']) ?>" placeholder="30"/>
+					<span><?php p($l->t('days to send a notification before the password expires'));?></span>
```
PVince81: this message sounds odd, maybe "number of days until expiration after which to send a notification"

@phil-davis any better ideas ?

tests/Db/OldPasswordMapperTest.php
```diff
@@ -57,46 +59,110 @@ protected function tearDown() {
...
+		$this->addInitialTestEntries($baseTime);
+
+
+		$passwordList = $this->mapper->getPasswordsAboutToExpire($baseTime + 14);
+		// last password change after the timestamp, so we shouldn't get any result
+		$this->assertCount(0, $passwordList);
+	}
+
+	public function testGetPasswordsAboutToExpireSomePassMarked() {
+		$baseTime = \time();
```
PVince81: maybe use a hard-coded value for deterministic tests ?

tests/Db/OldPasswordMapperTest.php
```diff
@@ -57,46 +59,110 @@ protected function tearDown() {
...
+
+		$uid = $this->testUIDs[1];
+		$latestPassword = $passwordList[1];
+		$this->assertSame("{$uid}testpass3", $latestPassword->getPassword());
+		$this->assertSame($uid, $latestPassword->getUid());
+		$this->assertLessThan($baseTime + 44, $latestPassword->getChangeTime());
+	}
+
+	public function testGetPasswordsAboutToExpireAllMarked() {
+		$baseTime = \time();
```
PVince81: same here

tests/HooksHandlerTest.php
```diff
@@ -84,8 +92,41 @@ protected function setUp() {
...
+			$this->manager,
+			$this->unConfigHandler
 		);
+
+		$this->manager->method('createNotification')
+			->will($this->returnCallback(function() {
+				$holder = [];
+				$mock = $this->createMock(INotification::class);
+				$mock->method('setApp')->will($this->returnCallback(function($app) use (&$holder, $mock) {
+					$holder['app'] = $app;
```
PVince81: there might be a better way, see https://github.com/owncloud/core/blob/master/apps/files_sharing/tests/NotifierTest.php#L145 and https://github.com/owncloud/core/blob/master/apps/files_sharing/tests/Service/NotificationPublisherTest.php#L101

I'm ok if you leave the tests as is

lib/Controller/SettingsController.php
```diff
@@ -92,9 +110,33 @@ public function getPriority() {
...
+
+	/**
+	 * Convert seconds to days. The value will always be rounded up,
+	 * so 1 second will be converted to 1 day
+	 * @param int $seconds the number of seconds to be converted
+	 * @return int the number of days in those seconds, rounded up
+	 */
+	private function secondsToDays($seconds) {
+		$floatDays = $seconds / (24 * 60 * 60);
+		return \intval(\ceil($floatDays));
```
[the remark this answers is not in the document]
phil-davis: The UI is only displaying integer number of days. So we have to do something here. When testing, I will understand that it will show 1 day, but underneath I actually set something like 300 seconds only.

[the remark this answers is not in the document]
jvillafanez: I'm not sure either.
If you set 12 hours instead of 1 day, you'll see 1 day instead of 0 days.

Maybe we should take the restrictive approach and show 0 days. This will send notifications before the expected time.

The good thing is that it's just a visualization problem in the UI if you set the value via occ. The notifications will be sent accordingly to the value stored in the DB (in seconds)

lib/Db/OldPasswordMapper.php
```diff
@@ -63,4 +63,39 @@ public function getLatestPassword($uid) {
...
+
+		if ($result === false) {
+			$info = \json_encode($stmt->erroInfo());
+			$message = "Cannot get the passwords that are about to expire. Error: {$info}";
+			\OCP\Util::writeLog('password_policy', $message, \OCP\Util::ERROR);
+			return false;
+		}
+
+		while ($row = $stmt->fetch()) {
+			$oldPasswords[] = OldPassword::fromRow($row);
```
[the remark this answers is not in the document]
jvillafanez: `yield` seems a better idea than the current one.
Running the job more often might not be a good idea: we'll need to rely on offset and limit, and some password might have changed in between messing with the expected order. We might not sent notifications for some users. We'll also need to hit the DB with the same query several times per day, instead of once or twice per day.

lib/Notifier.php
jvillafanez: 👍

phil-davis:
@jvillafanez @PVince81 can someone rebase this please?
At the moment, if I checkout this branch then I lose other recent fixes that are in master. That makes it difficult when testing.

tests/Db/OldPasswordMapperTest.php
jvillafanez: I don't think it's needed. The rest of the timestamps are based on whatever the `$baseTime` is, and we have control over the current time via the timeFactory (also based on the `$baseTime`). There won't be problems about this test failing randomly

phil-davis:
Settings UI issue:
1) Set ``days until user password expires`` to ``14`` and check the box.
2) Set ``days to send a notification before the password expires`` to ``30`` and check the box.
3) Press Save

Expected behavior:
You cannot set (2) greater than or equal to (1).
Something should ensure that  (2) < (1)

Actual behavior::
It saves OK - something bad is going to happen later! I don't even want to bother finding out what that is.

jvillafanez:
rebased

jvillafanez:
> It saves OK - something bad is going to happen later! I don't even want to bother finding out what that is.
it aborts the execution with a kind of cryptic message: https://github.com/owncloud/password_policy/pull/27/files#diff-779257764a9f14bbfce5338eb4e88ff0R83

lib/Db/OldPasswordMapper.php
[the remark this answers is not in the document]
butonic: @jvillafanez the subselect is what you actually need, right? 

In the result set you want the columns, necessary to build the entitiy, which is the outer select. You do a join on uid AND the change time to the subselect. Here you could use the id to match the exact password. However, the subselect already selects all the information you want. No need to join that to the table to get the other columns:
```sql
mysql> SELECT * FROM oc_user_password_history;
+-----+-------+----------------------------------------------------------------+-------------+
| id  | uid   | password                                                       | change_time |
+-----+-------+----------------------------------------------------------------+-------------+
|   1 | admin | 1|$2y$10$iJcpqCvfrm7RN/EBd4NPIOQrZvVr76RBrBI4DZm7pHSMVzs51/jXC |  1528802661 |
|   2 | admin | 1|$2y$10$K3v18hQGc2QppEAUwfIlVeT3v5I0WpY.Q5GG2GtiGg20r0pyd0fmS |  1529010840 |
|   3 | admin | 1|$2y$10$G5JjxOGGXT1PdcmFNf.p.evOIu.64dTeDOPEdTdRSTSSeuXI3ntkS |  1529010886 |
|   4 | admin | 1|$2y$10$SuAmaBeyA0LBXji54ekG.eMHywoqP1h573VbsYgwTndCMqN6aTYcS |  1529012884 |
|   5 | admin | 1|$2y$10$gKn4dZ60ALZU8DKFLzxCgeMIop5I.G6/.4PSzYZcY2Pb3vA57y4NS |  1529012941 |
|   6 | admin | 1|$2y$10$.5LFRjes2A0O061MRWxSGuzqbpOfqWS7EnlP45jKC8.LNYg1tymBq |  1529013064 |
|   7 | admin | 1|$2y$10$NVaealeCQTdfv5aU68RQ.OVbEQKUfJwlln0Mfp8Ac03p7iHqQtY06 |  1529013338 |
|   8 | admin | 1|$2y$10$SaFlqZSrolZs4wqGd7j/AOWvIBBVM8SfLfVCzpqg3ZfRdx55/Jqh2 |  1529013408 |
|   9 | admin | 1|$2y$10$cx/UbEsFFj0YQ3ICajrcPeCXum8AGwkiz08pf.ZSB1/wnQSH19WZS |  1529408997 |
|  10 | admin | 1|$2y$10$YdLFzYh36M12OQ6VllkwCeLePRQrXn/PJNl77OTHACz4BAIazXnNO |  1529409291 |
|  11 | admin | 1|$2y$10$n64oJTgbkDDuQtEKVw4/yOREctvjP1qreD5zPwR7bU.M.HV37I0gG |  1529409359 |
|  12 | admin | 1|$2y$10$rXQ5h9nSPUg9UVPhiLmPaOhkYxvX.YEU9/UIw5CDA.vgEcUeAfuYW |  1529409618 |
|  13 | admin | 1|$2y$10$H7q2D2JxgZ8CuuPqyBMqIOCxbwKl.8gSiE2vl.B5lFBDQSJ8Mcbva |  1529409667 |
|  14 | admin | 1|$2y$10$6bQRZnS0ezfJWsLKmbPX2.eyGiL6PM6d147kjMYnhSMvY6FG5oqj6 |  1529409986 |
|  15 | admin | 1|$2y$10$WaIcpl4Fm2BjxzEZx3.nvuG.vIvKkrUg0Vbu4M34ZpVTYc/AKAuNq |  1529410455 |
|  16 | admin | 1|$2y$10$awvxbG5wL54Zg6sYGgFpveskJ/bsjlPJVluKOtMe5r6/IgDHH3P2e |  1529410477 |
|  17 | admin | 1|$2y$10$UBEo7SutLVoQUQGysRk5seYvt8YUcl0M8Z3lP10m1KvuTtTDMT8i2 |  1529437059 |
|  18 | admin | 1|$2y$10$TWYrQpBkWMVJN45MmUi1KOaD.9gsRTUlQoYeaBpXdB/Xkuzb9aEli |  1529672819 |
|  19 | admin | 1|$2y$10$8QkkU1ZDwak0Ih2dIyTXW.1RmgwdYf9Es9bTD2UQuODAUpDc9xUTe |  1529672848 |
|  20 | admin | 1|$2y$10$Co8qKhNqf.ww9.BRzHbX0uh7E4p8XkPrctWZNxdMg3OSX4afImIQm |  1529673004 |
|  21 | admin | 1|$2y$10$l/bvtHoAzfgHX9O7VIsJiu1nUhQAd7y1WhAWZgpUm42fcDzYKtlne |  1529673041 |
|  22 | admin | 1|$2y$10$l1ToXrY7MgD2KkiZVDjYce.Q14m2vkWx7q1KlvloE4W3oZGlGN.t2 |  1529674775 |
|  23 | admin | 1|$2y$10$03WKH1TKscnOiqCoOTJoqOujnfJFlFSeWgObtSQj2nCh9DO99gfKK |  1529677007 |
|  24 | admin | 1|$2y$10$L/MDFrjGMnPm4nOiZ7mvuephhVUL1pvn1KsnMCPo3XwmCukB40R.K |  1529919562 |
|  25 | admin | 1|$2y$10$WRVgklX06FPQ9.gNGUTiLuMyTlinMw73DJZLqx3LJ2LlSBXB4WvvG |  1529919654 |
|  26 | admin | 1|$2y$10$VHt.R.lZX0PdYR86kJre.OgJCyaLXOgNjHAex6xXW/wKkOgRjnffK |  1529922722 |
|  27 | admin | 1|$2y$10$fXYr17rsqyztzu6HokqLpua/nm9fNf5mPZotAYKPLpvB2uU5C8gcq |  1529922783 |
|  28 | admin | 1|$2y$10$qR7w7vzae4XnZmWXOEhvEeGFqr.e7S6pKP0eEH4NCNHJHKiNwal4q |  1529928109 |
|  29 | admin | 1|$2y$10$x/be52NbGUdyqk2RtKihxOwr7QszeWVGHjRHcteiFwtNSRvXLH27K |  1529928317 |
|  30 | admin | 1|$2y$10$0/1NgwaFYkz0byBDP6FRauSesDuaa39KR.QlbbjXmDHX/LTDaHJU2 |  1529929051 |
|  31 | admin | 1|$2y$10$ncy2nAfKqopDD7xfL2xzQutajq1ULz7U7llKoYJA7f5kpJiyfISF6 |  1529929158 |
|  32 | admin | 1|$2y$10$Kojhu9ogf71pDODVoZAcwufBVAYYfZ0rskzTKGt.7u.0TcLKItn1K |  1529929495 |
|  33 | admin | 1|$2y$10$mQGjUuW52Si1SFpj8FaXfO4xFL6LJCLjrQm2WSahG2db0qw7vpjm. |  1529934553 |
|  34 | admin | 1|$2y$10$OFiETck8EABckSgbyKl6r.eOeWdd8WvHRR8OwEyVePN0ww9Gmer2m |  1530000377 |
|  35 | admin | 1|$2y$10$e8wPr0e.5qgv.rQ5BzKhRe8p5ErCe2kCIxWLhjsPcU3CzsBVVqs9i |  1530000452 |
|  36 | admin | 1|$2y$10$83uDOakjvWOhcBtv1QqnZu19AmMVSAd8Z7vHQ8yB326qBGdSS/Kty |  1530016008 |
|  37 | admin | 1|$2y$10$Igv7hYqcnNeBLASfr4UAxeYc1.5N/DJYqkHye/WBqLKVTcBIJhN6m |  1530019248 |
|  38 | admin | 1|$2y$10$ivCQvEIH2NS7KT.3nStpR.maBNcCMdOrPFWHvZqlzzUUr6Dk16of. |  1530087458 |
|  39 | admin | 1|$2y$10$pPNexxPypESD5MUobKKx..q4pbNzkhDSrb04fKvu29pJTb5cYaWtu |  1530087496 |
|  40 | admin | 1|$2y$10$wMAebL07zo3jfhY1g9f5KumVygs5HVIwnnYLoxTOilarc0oAO8GGS |  1530088064 |
| 215 | admin | 1|$2y$10$9DvIrgMhp16vS8PIIAiEL.9FgV03auKPWYzdnkLJ7ti0yyvo22SWG |  1530267611 |
| 216 | test  | 1|$2y$10$zfc50VR1AvRoAagYA/v94egSY9S1tmDzmK.stUWjP86ENem3KtO3G |  1530611134 |
| 217 | admin | 1|$2y$10$lCNQKQGGrfyag2Nc53w92eH/kuQEVxYX66RphyvoByE.nll.B0j6q |  1531385660 |
+-----+-------+----------------------------------------------------------------+-------------+
43 rows in set (0.00 sec)

mysql> SELECT `f`.`id`, `f`.`uid`, `f`.`password`, `f`.`change_time` FROM (   SELECT `uid`, max(`change_time`) AS `maxtime` FROM `oc_user_password_history` GROUP BY `uid`   ) AS `x` INNER JOIN `oc_user_password_history` AS `f` ON `f`.`uid` = `x`.`uid` AND `f`.`change_time` = `x`.`maxtime`   WHERE `f`.`change_time` < 1540019248;
+-----+-------+----------------------------------------------------------------+-------------+
| id  | uid   | password                                                       | change_time |
+-----+-------+----------------------------------------------------------------+-------------+
| 217 | admin | 1|$2y$10$lCNQKQGGrfyag2Nc53w92eH/kuQEVxYX66RphyvoByE.nll.B0j6q |  1531385660 |
| 216 | test  | 1|$2y$10$zfc50VR1AvRoAagYA/v94egSY9S1tmDzmK.stUWjP86ENem3KtO3G |  1530611134 |
+-----+-------+----------------------------------------------------------------+-------------+
2 rows in set (0.00 sec)

mysql> SELECT `id`, `uid`, `password`, max(`change_time`) AS `change_time` FROM `oc_user_password_history` WHERE `change_time` < 1540019248 GROUP BY `uid`;
+-----+-------+----------------------------------------------------------------+-------------+
| id  | uid   | password                                                       | change_time |
+-----+-------+----------------------------------------------------------------+-------------+
|   1 | admin | 1|$2y$10$iJcpqCvfrm7RN/EBd4NPIOQrZvVr76RBrBI4DZm7pHSMVzs51/jXC |  1531385660 |
| 216 | test  | 1|$2y$10$zfc50VR1AvRoAagYA/v94egSY9S1tmDzmK.stUWjP86ENem3KtO3G |  1530611134 |
+-----+-------+----------------------------------------------------------------+-------------+
2 rows in set (0.00 sec)
```

Can you double check if I misunderstand the query you built? You should be able to get the same results with this.

The indexes that need to be used will vary, depending on the number of passwords per user in the table. 

`yield` would be great, but does not add too much benefit. even passwords 300k users should fit in memory ...

jvillafanez:
squashed the last 2 commits and included a fix for the failing test.

lib/Db/OldPasswordMapper.php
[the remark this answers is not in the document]
jvillafanez:
```
mysql> SELECT `id`, `uid`, `password`, max(`change_time`) AS `change_time` FROM `user_password_history` WHERE `change_time` < 1540019248 GROUP BY `uid`;
ERROR 1055 (42000): Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'owncloud.user_password_history.id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
```
I expect something similar with postgresql.

lib/Jobs/PasswordExpirationNotifierJob.php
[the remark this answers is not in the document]
PVince81: occ command allows past and future:
`occ user:expire-password user1 "5 days"`
`occ user:expire-password user1 -- '-5 days'`

@jvillafanez would this break here ?

PVince81:
damn, created a conflict with parallel merge

lib/Db/OldPasswordMapper.php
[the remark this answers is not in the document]
butonic: I see. Then move the mtime comparison into the subselect. Hm or we do a left joun to get rid of the derived table:
```sql
mysql> SET SESSION sql_mode='ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION';
Query OK, 0 rows affected (0.00 sec)

mysql> SELECT p.* FROM `oc_user_password_history` p LEFT OUTER JOIN `oc_user_password_history` p2 ON p.uid = p2.uid AND p.change_time < p2.change_time WHERE p2.id IS NULL;
+-----+-------+----------------------------------------------------------------+-------------+
| id  | uid   | password                                                       | change_time |
+-----+-------+----------------------------------------------------------------+-------------+
| 216 | test  | 1|$2y$10$zfc50VR1AvRoAagYA/v94egSY9S1tmDzmK.stUWjP86ENem3KtO3G |  1530611134 |
| 217 | admin | 1|$2y$10$lCNQKQGGrfyag2Nc53w92eH/kuQEVxYX66RphyvoByE.nll.B0j6q |  1531385660 |
+-----+-------+----------------------------------------------------------------+-------------+
2 rows in set (0.00 sec)

mysql> EXPLAIN SELECT p.* FROM `oc_user_password_history` p LEFT JOIN `oc_user_password_history` p2 ON p.uid = p2.uid AND p.change_time < p2.change_time WHERE p2.id IS NULL;
+----+-------------+-------+------+-------------------------------------------------------------------+--------------------+---------+-------------+------+--------------------------------------+
| id | select_type | table | type | possible_keys                                                     | key                | key_len | ref         | rows | Extra                                |
+----+-------------+-------+------+-------------------------------------------------------------------+--------------------+---------+-------------+------+--------------------------------------+
|  1 | SIMPLE      | p     | ALL  | NULL                                                              | NULL               | NULL    | NULL        |   43 |                                      |
|  1 | SIMPLE      | p2    | ref  | pp_uid_index,pp_mtime_index,pp_uid_mtime_index,pp_mtime_uid_index | pp_uid_mtime_index | 258     | core1.p.uid |    1 | Using where; Using index; Not exists |
+----+-------------+-------+------+-------------------------------------------------------------------+--------------------+---------+-------------+------+--------------------------------------+
2 rows in set (0.00 sec)
```
As you can see I added another index `pp_uid_mtime_index` with `CREATE INDEX pp_uid_mtime_index ON oc_user_password_history (uid, change_time);` and the query can use it: `Using where; Using index; Not exists`

templates/admin.php
phil-davis: "send a notification this many days before the password expires"
If it is OK to use this style of grammar.
Forcing the sentence so that it begins with the number "30" makes the English difficult, whatever you do.

lib/Db/OldPasswordMapper.php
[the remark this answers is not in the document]
butonic: Hm ... it really depends on the dbms which query is more efficient. go with the current code. In any case this wont be a performance issue.

PVince81:
settings UI to be done later: https://github.com/owncloud/password_policy/issues/41

PVince81:
damn, I broke the tests by changing the texts...

PVince81:
SQL to be addressed separately: https://github.com/owncloud/password_policy/issues/42

Issue #30, opened 2018-07-10
Admin can expire password of LDAP user


phil-davis:
Steps to reproduce:
1. setup LDAP on your system - e.g. https://github.com/owncloud/core/wiki/How-to-test-user_ldap
2. from the command line, attempt to expire the password of an LDAP user:
```
./occ user:expire-password zombie5
The password for zombie5 is set to expire on 2018-07-09 04:01:13 UTC.
```
3. Try to login as ``zombie5``

Expected behavior:
The ``user:expire-password`` command should fail for an LDAP user.

Actual behavior:
The password gets expired (from the local point of view of ownCloud).
The user ``zombie5`` is taken to the password change page when they try to login to the webUI.
If they fill in old and new passwords..., then "Internal Server Error" is reported.

phil-davis:
PR  to fix #31

phil-davis:
PR merged.

Issue #29, opened 2018-07-10
Attempts to change LDAP user passwords give password policy messages


phil-davis:
Steps to reproduce:
1. setup LDAP on your system - e.g. https://github.com/owncloud/core/wiki/How-to-test-user_ldap
2. as an admin, on the webUI, attempt to change the password of an LDAP user
- you get messages about the password being too short...
- when you enter a password that meets the policy, then you get ``Unable to change password`` 

3. from the command line, attempt to reset the password of an LDAP user:
```
./occ user:resetpassword zombie5
Enter a new password: 
Confirm the new password: 

In Length.php line 33:
                                                                  
  The password is too short. At least 8 characters are required.  
                                                                  

user:resetpassword [--password-from-env] [--] <user>
```

If you enter a password that meets the policy, then you get:
```
./occ user:resetpassword zombie5
Enter a new password: 
Confirm the new password: 
Error while resetting password!
```

The password policy hooks are "getting in first" and misleading you into thinking that maybe you can have success if you meet the policy requirements in the message.

Nothing is broken, but it would be a nicer user experience if:

1) ``password_policy`` hooks do not get invoked for non-local users (or their code understands and does nothing, or they are invoked after the code that emits ``Unable to change password`` so that actually that error message gets emitted first or...)

2) Messages like ``Unable to change password`` and ``Error while resetting password!`` gave a better clue about why - that the password is managed by some other back-end authentication system.

Item 2 is not the responsibility of ``password_policy``, I am just mentioning it here while I think of it.

PVince81:
from my understand you shouldn't even be able to change the password at all from OC for LDAP users, so we should remove the field from the UI.

for the CLI part, would need to detect that changing password is not supported for that user's backend and abort directly with an error message

phil-davis:
> so we should remove the field from the UI

Which will be a core change. There will still be a "password" column, because the list of users on the Users Page can be a mix of LDAP and local and... I guess the password field would just be hidden for non-local users, leaving some blank space in the column.

phil-davis:
> for the CLI part, would need to detect that changing password is not supported for that user's backend and abort directly with an error message

done in PR #31

PVince81:
to be followed up in https://github.com/owncloud/user_management/issues/42

PR #31, opened 2018-07-10
Do not allow passwords to be expired for non-local users

phil-davis:
Fixes issue #30

### DIFF
--- a/lib/Command/ExpirePassword.php
+++ b/lib/Command/ExpirePassword.php
@@ -78,14 +78,21 @@ class ExpirePassword extends Command {
 	protected function execute(InputInterface $input, OutputInterface $output) {
 		$uid = $input->getArgument('uid');
 
-		$exists = $this->userManager->userExists($uid);
-		if($exists === false) {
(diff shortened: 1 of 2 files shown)


lib/Command/ExpirePassword.php
```diff
@@ -88,6 +90,11 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 			return 67;
 		}
 
+		if (!$user->canChangePassword()) {
+			$output->writeln("<error>Passwords cannot be expired for user: $uid</error>");
```
PVince81: should we add a reason: because the user is in a backend that doesn't support password change ?

phil-davis: Why not.
I mentioned this sort of thing in https://github.com/owncloud/password_policy/issues/29#issue-339683422 (item 2 at the bottom of the post). In core, it would be nice to make some more friendly messages when password change is attempted for these type of users.

phil-davis:
Optional comment addressed. "cannot be expired" message expanded.

Issue #25, opened 2018-07-09
When "force change on first login" is not set, user password never expires


phil-davis:
Steps to reproduce

1. As admin, install and enable the ``password_policy`` app
2. Set a user password policy of ``xx`` days until user password expires
3. Do not set "force change on first login"
4. As admin, add a user account ``user123``
5. Login as ``user123`` and make sure things are "normal for you". You are not asked to change your password.
6. Wait ``xx`` days
7. Login as ``user123``

Expected behavior:
- the user is taken to the password change page

Actual behavior:
- the password continues to work

phil-davis:
To test this, I use my branch in PR #23 which shortens the password expiry to be in seconds, not days.
That works fine for me in ordinary testing.

phil-davis:
I suspect that this also has another somewhat less annoying side effect:
1) Policy is to change passwords every 90 days
2) Have a user that changed their password 80 days ago
3) The user is in a mess, forgotten their password and somehow cannot reset their own password. They sit with the sysadmin/helpdesk who changes their password from an admin account
4) 10 days later they get forced to change their password (because the password change by admin is not remembered as a user password change)

PVince81:
Estimate: 0.5md

Proposed technical solution:

- Detect password change through user creation and password change hooks, and store the timestamp in the database.

pmaier1:
Yep, release blocker.

From https://github.com/owncloud/password_policy/issues/24#issuecomment-403464859

> As discussed in a chat, let's use the "date when the policy got activated" for every user where no record exists.

You would then have a point in time. But for a policy that was activated two years ago, this would mean every new user would need to change their password on first login, though. That's tricky. 

> Proposed technical solution:
Detect password change through user creation and password change hooks, and store the timestamp in the database.

Yep, so for every new user (and when pw changed by admin) there will be a record. I like the idea. @butonic please confirm/comment.

butonic:
If the idea is to store the timestamp in the old passwords table, then +1

phil-davis:
I agree, populating ``user_password_history`` with the initial password that the admin gives seems quite reasonable, and will ensure all new users get off to a good start.

butonic:
if a password change by the admin does not add a history entry that is a bug

sharidas:
Below are my observations in the environment master branches of core + password_policy
- Create `admin` user and enable password_policy app.
- Apply https://github.com/owncloud/password_policy/pull/23 on top of the master branch of password_policy
[...]

PR #34, opened 2018-07-10
Create a new entry in user_password_history when a new user is created

jvillafanez:
Fix https://github.com/owncloud/password_policy/issues/25 by creating a new entry when the user is created

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -42,6 +42,10 @@ $eventDispatcher->addListener(
 	[$handler, 'generatePassword']
 );
 $eventDispatcher->addListener(
+	'user.aftercreate',
+	[$handler, 'savePasswordForCreatedUser']
+);
+$eventDispatcher->addListener(
 	'user.aftersetpassword',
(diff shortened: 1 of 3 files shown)


phil-davis:
Works via all 3 user-creation methods, webUI, API and occ command.
Good stuff.

phil-davis:
But the tes fails:
```
There was 1 error:
1) OCA\PasswordPolicy\Tests\HooksHandlerTest::testSavePasswordForCreatedUser
InvalidArgumentException: Argument "user" not found.
```

PVince81:
we also need one when the admin changes a user's password

sharidas:
> we also need one when the admin changes a user's password

This is already addressed here https://github.com/owncloud/password_policy/blob/master/appinfo/app.php#L44-L47
This event is dispatched from here -> https://github.com/owncloud/core/blob/master/lib/private/User/User.php#L281

phil-davis:
> we also need one when the admin changes a user's password
and it works.

PR #35, opened 2018-07-10
Check password type + allow empty public password

PVince81:
- Pass along new password "type" argument which can be "user", "guest" or "public" from "validatePassword" event.
- Allow empty public link passwords even when rules are set

In the future this will allow easier splitting of rules between "user password" and "public link password".

The PR as is will not do anything and requires the following core changes to work:
- [ ] TODO: core PR

### DIFF
--- a/lib/Engine.php
+++ b/lib/Engine.php
@@ -105,9 +105,10 @@ class Engine {
 	/**
 	 * @param string $password
 	 * @param string $uid
+	 * @param string $type
 	 * @throws PolicyException
 	 */
-	public function verifyPassword($password, $uid = null) {
+	public function verifyPassword($password, $uid = null, $type = 'user') {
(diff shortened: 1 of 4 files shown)


lib/Engine.php
```diff
@@ -105,9 +105,16 @@ public function generatePassword() {
 	/**
 	 * @param string $password
 	 * @param string $uid
+	 * @param string $type
 	 * @throws PolicyException
 	 */
-	public function verifyPassword($password, $uid = null) {
+	public function verifyPassword($password, $uid = null, $type = 'user') {
+		// skip policy when no password is specified
```
phil-davis: skip policy when no password is specified for a public link

PVince81:
Superseded by https://github.com/owncloud/password_policy/pull/36

PR #36, opened 2018-07-10
Check password type based on event name + allow empty password for public links

PVince81:
- Pass along new password "type" argument which can be "user", "public" based from the namespace of the "validatePassword" event
- Allow empty public link passwords even when rules are set

No core changes required.


PVince81:
Tested against OC master: with a policy rule min 8 chars:
- [x] TEST: creating public link without password works
- [x] TEST: creating public link with password "a" fails with error as the password is too short

phil-davis:
@PVince81 you need to "sign" the CLA

PVince81:
fixed comment based on review from former PR

PR #38, opened 2018-07-11
Adapt description to incorporate recently added new features

pmaier1:
@settermjd @phil-davis I think this still has some room for improvement. Maybe you can help a little with wording magic. 

@PVince81 and/or @butonic maybe also have some feedback.

THX!


appinfo/info.xml
```diff
@@ -6,9 +6,12 @@
...
+Password history and expiration policies are supplements that allow IT to establish a level of password security that can comply with corporate guidelines of all sorts. The provided tools enable administrators to granularly choose their desired security level. At this point it is important to keep in mind that high levels of security might sacrifice usability and come at the expense of user experience. For this reason it is highly recommended to carefully decide on the hurdles that are put upon users in order to maintain and optimize user adoption and satisfaction.
+
+Administrators find the configuration options in the 'Security' section of the ownCloud administration settings panel. The respective policies are designed for local user accounts, not for user accounts imported from LDAP or other user backends as these provide own mechanisms. For more information and recommendations when deploying policies in an existing ownCloud, please consult the [ownCloud Documentation](https://doc.owncloud.com/server/latest/admin_manual/configuration/server/security/password-policy.html).</description>
```
phil-davis: ``as these provide own mechanisms`` ->
``as these provide their own mechanisms``
("for managing these things" is implied, and I don't think needs to be said)

phil-davis:
There is no reference here to the NIST guidelines, i.e. that actually these days forcing password expiry, for example, is no longer seen as a best practice. But the option is there so that if a company still requires it, then the app can do it.
I guess that will be in the linked documentation.

PVince81:
@pmaier1 can you update ?

pmaier1:
Done. Thx @phil-davis!

PR #46, opened 2018-07-12
New password must not match current one when resetting

PVince81:
When forcing existing users to change their password and no entry exists
in the history table, we must still prevent the user to set the password
to the same as the current one.

Found this bug and fixed it directly.

### DIFF
--- a/lib/Controller/PasswordController.php
+++ b/lib/Controller/PasswordController.php
@@ -145,6 +145,13 @@ class PasswordController extends Controller implements IAccountModuleController 
 			);
 		}
 
+		if ($new_password === $current_password) {
+			return $this->createPasswordTemplateResponse(
+				$redirect_url,
+				$this->l10n->t('Password must be different than the old password.')
+			);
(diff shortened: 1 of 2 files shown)


PVince81:
### Steps:

1. Enable password policy app
1. Create user "user1" with password "test"
1. Enable rule "force first login"
1. Log out
1. Login as "user1"
1. Type "test", "test" and "test" in the fields

### Expected (after fix)
Password not accepted

### Actual (before fix)
Password accepd

phil-davis:
Well found! I am a bit surprised that this workflow could happen any more, because now when you create ``user1`` the ``password_policy`` app is supposed to have a hook on the user create, and put that first admin-created user password into the history for the user. So I expected that it would be discovered in the password history and the match refused.

Anyway, it is nice to report specifically that the old and new password were the same.

PVince81:
@phil-davis not sure either. Could be due to switching branches or fancy combination of options and bad luck

Issue #47, opened 2018-07-12
Notification not created when using occ user:expire-password command


PVince81:
### Steps

1. Login as admin
1. Enable rule with 20 days until user password expires
1. Enable rule with 5 days after password expires, send notification
1. Create a user "user1" with password "test"
1. Login as "user1"
1. Run `occ user:expire-password user1 -- "-40 days"`
1. Run cron after resetting oc_jobs timetamps (the usual way to test with cron.php)
1. Refresh the page

### Expected result
Notification appears about expired password.
If a smaller value was selected, a notification about password is about to expire

### Actual result
No notification at all

### Version
Current master (2.0.0)

Looking at the code it seems that the mechanism for actual expiration is done based on the user preference "forcePasswordChange". However the notification mechanism is checking the timestamp of the last password in the password history table.

It seems that the notification code should rather look at the "forcePasswordChange" value as this is the one influenced by the last password changes. I have the feeling that there were misunderstandings during the implementation and review.

@jvillafanez please adjust the code to select based on preferences.
I seem to remember about problems when parsing the string value to int with Oracle as the preference is a CLOB and maybe using the history table is a workaround ?

In the latter case we might need to adjust the occ command to also insert a dummy password entry into the table with the same timestamp as "forcePasswordChange".

I'm not sure yet about how important this is for the release. Based on https://github.com/owncloud/password_policy/issues/24 for existing users it was said that one needs to use the occ command. So it might be best to make it work. The faster fix might be inserting a dummy password.

@butonic does that make sense ?

PVince81:
@pmaier1 this is the reason I couldn't demo the notification. In the past I used another hack where I touched the history table but forgot at the time we talked.

@pmaier1 any thoughts on criticality considering the timeline ?

pmaier1:
Hmm, well, this is a UX thing. Especially for existing instances that impose policies users will not receive a notification.
- won't know that password has expired until they log in again
- when not using oauth clients might throw errors about wrong password? what if we combine this with the brute-force protection app?

If these two statements are true, yes, this is at least a p2. Not sure about the effort. I actually can't defer the planned release date.

jvillafanez:
It will depend on how we want to deal with this:
* We can send an specific notification for this case without worrying to touch the cron job. I'd expect the user to reset the password before getting more notifications from the cron job. Kind of bad as the user might receive notification from this command and the cron job.
[...]

PR #51, opened 2018-07-13
Send an expired notification when the admin expires a password

jvillafanez:
Fix https://github.com/owncloud/password_policy/issues/47

Notification will be sent on the next cron job run.

### DIFF
--- a/lib/Command/ExpirePassword.php
+++ b/lib/Command/ExpirePassword.php
@@ -110,14 +110,19 @@ class ExpirePassword extends Command {
 			return 1;
 		}
 
-		$date = new \DateTime();
-		$date->setTimezone(new \DateTimeZone('UTC'));
-		$date->setTimestamp($this->timeFactory->getTime());
-		$date->modify($input->getArgument('expiredate'));
+		$expireDate = new \DateTime();
+		$expireDate->setTimezone(new \DateTimeZone('UTC'));
+		$expireDate->setTimestamp($this->timeFactory->getTime());
+		$expireDate->modify($input->getArgument('expiredate'));
+
+		$oldDate = new \DateTime();
+		$oldDate->setTimezone(new \DateTimeZone('UTC'));
+		$oldDate->setTimestamp($this->timeFactory->getTime());
+		$oldDate->modify($input->getArgument('expiredate'));
 
 		if ($this->config->getAppValue('password_policy', 'spv_user_password_expiration_checked', false) === 'on') {
 			$delta = $this->config->getAppValue('password_policy', 'spv_user_password_expiration_value', 90);
-			$date->modify("-$delta days");
+			$oldDate->modify("-$delta days");
 		}
 
 		$this->config->deleteUserValue(
@@ -131,12 +136,12 @@ class ExpirePassword extends Command {
 		$oldPassword = new OldPassword();
 		$oldPassword->setUid($uid);
 		$oldPassword->setPassword(OldPassword::EXPIRED);
-		$oldPassword->setChangeTime($date->getTimestamp());
+		$oldPassword->setChangeTime($oldDate->getTimestamp());
 		$this->mapper->insert($oldPassword);
 
 		// show expire date if it was given
 		if ($input->hasArgument('expiredate')) {
-			$output->writeln("The password for $uid is set to expire on ". $date->format('Y-m-d H:i:s T').'.');
+			$output->writeln("The password for $uid is set to expire on ". $expireDate->format('Y-m-d H:i:s T').'.');
 		}
 		return 0;
 	}
--- a/lib/Command/ExpirePassword.php
+++ b/lib/Command/ExpirePassword.php
@@ -30,6 +30,7 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use OCP\AppFramework\Utility\ITimeFactory;
 
 
 class ExpirePassword extends Command {
@@ -45,17 +46,20 @@ class ExpirePassword extends Command {
 	/**
 	 * @param IConfig $config
 	 * @param IUserManager $userManager
+	 * @param ITimeFactory $timeFactory
 	 * @param OldPasswordMapper $mapper
 	 */
 	public function __construct(
 		IConfig $config,
 		IUserManager $userManager,
+		ITimeFactory $timeFactory,
 		OldPasswordMapper $mapper
 	) {
 		parent::__construct();
 		$this->config = $config;
 		$this->userManager = $userManager;
 		$this->mapper = $mapper;
+		$this->timeFactory = $timeFactory;
 	}
 
 	protected function configure() {
@@ -106,10 +110,12 @@ class ExpirePassword extends Command {
 			return 1;
 		}
 
-		$date = new \DateTime($input->getArgument('expiredate'));
+		$date = new \DateTime();
 		$date->setTimezone(new \DateTimeZone('UTC'));
+		$date->setTimestamp($this->timeFactory->getTime());
+		$date->modify($input->getArgument('expiredate'));
 
-		if ($this->config->getAppValue('password_policy', 'spv_user_password_expiration_checked', 'no') === 'on') {
+		if ($this->config->getAppValue('password_policy', 'spv_user_password_expiration_checked', false) === 'on') {
 			$delta = $this->config->getAppValue('password_policy', 'spv_user_password_expiration_value', 90);
 			$date->modify("-$delta days");
 		}
--- a/lib/Command/ExpirePassword.php
+++ b/lib/Command/ExpirePassword.php
@@ -24,6 +24,8 @@ namespace OCA\PasswordPolicy\Command;
 
 use OCP\IConfig;
 use OCP\IUserManager;
+use OCA\PasswordPolicy\Db\OldPasswordMapper;
+use OCA\PasswordPolicy\Db\OldPassword;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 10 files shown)


lib/Command/ExpirePassword.php
```diff
@@ -106,6 +116,19 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 			$value
 		);
 
+		// add a dummy password in the user_password_history so the cron job
+		// can notify about the expiration of the password. We'll use "dummy"
+		// as password hash (we won't hash "dummy")
+		$latestPassword = $this->mapper->getLatestPassword($uid);
+		if ($latestPassword->getPassword() !== 'dummy') {
```
PVince81: we could use a constant maybe

lib/Command/ExpirePassword.php
```diff
@@ -106,6 +116,19 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
+			// insert dummy only if it isn't the last
+			$oldPassword = new OldPassword();
+			$oldPassword->setUid($uid);
+			$oldPassword->setPassword('dummy');
+			$oldPassword->setChangeTime(\max($date->getTimestamp(), $latestPassword->getChangeTime() + 1));
```
PVince81: The notification would not work if we set a negative time with the occ command as the `max` will always take the timestamp of the latest password. I think this is ok as a negative value would trigger immediate expiry, and immediate expiry notification. This is ok.

If an admin runs the occ command twice with two different values, you must still insert a new "dummy" entry to have the most recent timestamp value. This would also keep track indirectly in history table about how many times an admin has changed the expiration time.

PVince81: - [ ] doubt about `$date->getTimestamp()`. if through occ we say "password must expire in 5 days", we shouldn't set the timestamp in the future but to "future date - days to expire". This way the expiration job should have accurate value.

@jvillafanez did you test the current behavior on your PR and double check what "days" value you see in the notification ?

PVince81: history timestamp in the history table is supposed to be "$expireDate - $daysUntilExpiry" so that the notification can show "your password will expire in $daysUntilExpiry" notification.

now how to calculate "$daysUntilExpiry" here in a way that makes it work... should we use the existing rule ?
maybe `$daysUntilExpiry = time() - $date->getTimestamp()` ?

PVince81:
let's go with https://github.com/owncloud/password_policy/issues/47#issuecomment-404795581 instead. 

the dummy history entry doesn't save us time but makes us waste even more... can of worms.

lib/Command/ExpirePassword.php
PVince81: this is the formula: `$changeTime = $date->getTimestamp() - $passwordExpiresAfterDaysRule`

PVince81:
- [ ] don't set oc_preferences flag from occ command (as discussed with @butonic)

lib/Command/ExpirePassword.php
PVince81: - [ ] regarding `\max` => don't insert dummy entry if the previous entry is more recent than the calculated `$changeTime`

PVince81:
- [ ] rename "dummy" to "expired"
- [ ] order by id instead of change_time
    - [ ] bkg job
    - [ ] password history

PVince81:
- [ ] double check that all usages of `getLatestPassword` and `getOldPasswords` can deal with dummy entry.

jvillafanez:
diff changes to sort by id instead of change_time (based on this branch):
```
diff --git a/lib/Db/OldPasswordMapper.php b/lib/Db/OldPasswordMapper.php
index 3050fdb..98223dd 100644
--- a/lib/Db/OldPasswordMapper.php
+++ b/lib/Db/OldPasswordMapper.php
@@ -42,7 +42,7 @@ class OldPasswordMapper extends Mapper {
 		$qb->select('*')
 			->from('user_password_history')
 			->where($qb->expr()->eq('uid', $qb->createNamedParameter($uid)))
-			->orderBy('change_time', 'desc')
+			->orderBy('id', 'desc')
 			->setMaxResults($length);
 		$result = $qb->execute();
 		$rows = $result->fetchAll();
@@ -75,8 +75,8 @@ class OldPasswordMapper extends Mapper {
 	 */
 	public function getPasswordsAboutToExpire($maxTimestamp) {
 		$query = "SELECT `f`.`id`, `f`.`uid`, `f`.`password`, `f`.`change_time` FROM (";
-		$query .= "SELECT `uid`, max(`change_time`) AS `maxtime` FROM `*PREFIX*user_password_history` GROUP BY `uid`";
-		$query .= ") AS `x` INNER JOIN `*PREFIX*user_password_history` AS `f` ON `f`.`uid` = `x`.`uid` AND `f`.`change_time` = `x`.`maxtime`";
+		$query .= "SELECT `uid`, max(`id`) AS `maxid` FROM `*PREFIX*user_password_history` GROUP BY `uid`";
+		$query .= ") AS `x` INNER JOIN `*PREFIX*user_password_history` AS `f` ON `f`.`uid` = `x`.`uid` AND `f`.`id` = `x`.`maxid`";
 		$query .= " WHERE `f`.`change_time` < ? OR `password` = 'dummy'";
 
 		$stmt = $this->db->prepare($query);
```

phil-davis:
I would think that:
``getLatestPassword`` - is just for getting the latest "real" password hash value - maybe make just a ``getLatestPasswordHash`` ?

``getOldPasswords`` - is just for getting the list of "real" old passwords, for checking password matches.

Neither of these ever need to return the "force-expire" entries.

And have a:
``getLatestEffectivePasswordChange`` that returns the top "order by id" entry, which is the date-time-stamp to use when calculating password expiry/notification.

PVince81:
(pending commit on @butonic's machine, we're testing)

- [ ] TEST: notification is sent out without dummy entry
- [ ] TEST: notification is sent out with dummy entry
- [ ] TEST: expiration still works without dummy entry
- [ ] TEST: expiration still works with dummy entry

PVince81:
ok, looks good to me so far

I'll update the unit tests.

PVince81:
Adjusted unit tests, they all pass for me locally.

I hope codecov can report something...

PVince81:
ohh, codecov is happy.

@butonic @jvillafanez please review my changes.

if all ok, please squash.

then next step is integrate https://github.com/owncloud/password_policy/pull/55

lib/Db/OldPasswordMapper.php
```diff
@@ -47,7 +47,7 @@ public function getOldPasswords($uid, $length, $excludeForceExpired = true) {
 			->orderBy('id', 'desc')
 			->setMaxResults($length);
 		if ($excludeForceExpired) {
-			$qb->andWhere($qb->expr()->neq('password', OldPassword::EXPIRED));
+			$qb->andWhere($qb->expr()->neq('password', $qb->expr()->literal(OldPassword::EXPIRED)));
```
butonic: ok, so you were right ...

lib/Jobs/PasswordExpirationNotifierJob.php
```diff
@@ -97,15 +97,16 @@ protected function run($arguments) {
...
-				$this->logger->debug("password timestamp for {$passInfo->getUid()}: {$passInfo->getChangeTime()}; elapsed time: {$elapsedTime} -> EXPIRED", ['app' => 'password_policy']);
+				$adminPart = '; initiated by admin';
+			}
+
+			if ($elapsedTime >= $expirationTime) {
+				$this->logger->debug("password timestamp for {$passInfo->getUid()}: {$passInfo->getChangeTime()}; elapsed time: {$elapsedTime} -> EXPIRED${adminPart}", ['app' => 'password_policy']);
 				$this->sendPassExpiredNotification($passInfo, $expirationTime);
 			} elseif ($elapsedTime >= $notifyAfter) {
-				$this->logger->debug("password timestamp for {$passInfo->getUid()}: {$passInfo->getChangeTime()}; elapsed time: {$elapsedTime} -> NOTIFY", ['app' => 'password_policy']);
+				$this->logger->debug("password timestamp for {$passInfo->getUid()}: {$passInfo->getChangeTime()}; elapsed time: {$elapsedTime} -> NOTIFY${adminPart}", ['app' => 'password_policy']);
```
butonic: ok, so you log that the admin triggered the notification as well

tests/Rules/PasswordExpiredTest.php
```diff
@@ -112,6 +112,27 @@ public function testWrongPasswordLogsWarning() {
...
+			->willReturn($password);
+		$this->hasher
+			->expects($this->never())
+			->method('verify');
+		$this->timeFactory
+			->expects($this->once())
+			->method('getTime');
+		$this->logger
+			->expects($this->never())
+			->method('warning');
```
butonic: ack

lib/Command/ExpirePassword.php
```diff
@@ -95,17 +110,30 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 			return 1;
 		}
 
-		$date = new \DateTime($input->getArgument('expiredate'));
+		$date = new \DateTime();
```
butonic: changed to use the time factory to make testing relative dates possible, nice

phil-davis:
> then next step is integrate #55

done - I adjusted unit tests. They need to expect 2 different dates, the expire date as reported to the user in the command output and the history date that is stuffed in the ``user_password_history`` table.

lib/Command/ExpirePassword.php
```diff
@@ -95,20 +113,38 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
-		$value = $date->format('Y-m-d\TH:i:s\Z'); // ISO8601 with Zulu = UTC
+		$expireDate = new \DateTime();
+		$expireDate->setTimezone(new \DateTimeZone('UTC'));
+		$expireDate->setTimestamp($this->timeFactory->getTime());
+		$expireDate->modify($input->getArgument('expiredate'));
+
+		$oldDate = new \DateTime();
+		$oldDate->setTimezone(new \DateTimeZone('UTC'));
+		$oldDate->setTimestamp($this->timeFactory->getTime());
+		$oldDate->modify($input->getArgument('expiredate'));
```
phil-davis: Someone could extract method to refactor these 2 blocks that make ``$expireDate`` and ``$oldDate``, or is there magic in PHP to really copy/clone and object to make ``$oldDate`` a real copy of ``$expireDate`` then modify ``$oldDate`` below.

PVince81:
Great team work here, passing the ball around. Thanks a bunch!
Merging...

lib/Db/OldPasswordMapper.php
PVince81: yep, this blew up during unit testing...

PR #55, opened 2018-07-13
Report correct expire date for expire-password command

phil-davis:
The command was doing stuff like:
```
./occ user:expire-password u222
The password for u222 is set to expire on 2018-04-13 15:45:48 UTC.

./occ user:expire-password u222 "2018-07-20"
The password for u222 is set to expire on 2018-04-21 00:00:00 UTC.
```
reporting the date 90 days before the expiry date.

After this change:
```
./occ user:expire-password u222 "2018-07-21 00:01:00"
The password for u222 is set to expire on 2018-07-21 00:01:00 UTC.
```
and the ``user_password_history`` table has a timestamp back 90 days from that, as it should.

### DIFF
--- a/lib/Command/ExpirePassword.php
+++ b/lib/Command/ExpirePassword.php
@@ -110,14 +110,19 @@ class ExpirePassword extends Command {
 			return 1;
 		}
 
-		$date = new \DateTime();
-		$date->setTimezone(new \DateTimeZone('UTC'));
-		$date->setTimestamp($this->timeFactory->getTime());
-		$date->modify($input->getArgument('expiredate'));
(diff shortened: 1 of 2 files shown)


PVince81:
holding this until https://github.com/owncloud/password_policy/pull/51 is merged as it brings in more unit tests to validate

phil-davis:
1) Resolved conflicts and fixed unit tests so they know to expect a different reported time as the command result from the history time that is put in the ``user_password_history`` table.

2) Added the definition of ``timeFactory`` var, which PHPstorm shouted at me about.

PR #60, opened 2018-07-17
Return a constant not a magic number if the user is null

settermjd:
This PR refactors the return of the magic (though documented) number "67" as a class constant. While the return statement, when the user is null, is clearly documented, I believe that using a constant instead of a magic number makes the code clearer and more maintainable.

### DIFF
--- a/tests/Command/ExpirePasswordTest.php
+++ b/tests/Command/ExpirePasswordTest.php
@@ -68,12 +68,14 @@ class ExpirePasswordTest extends TestCase {
 			->with('not-existing-uid')
 			->willReturn(null);
 
-		$this->commandTester->execute([
+		$returnCode = $this->commandTester->execute([
 			'uid' => 'not-existing-uid',
 			'expiredate' => '2018-06-28 10:13 UTC'
-			]);
+		]);
 		$output = $this->commandTester->getDisplay();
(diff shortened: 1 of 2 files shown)


tests/Command/ExpirePasswordTest.php
```diff
@@ -68,12 +68,14 @@ public function testExpirePasswordUserNotExisting() {
 			->with('not-existing-uid')
 			->willReturn(null);
 
-		$this->commandTester->execute([
-			'uid' => 'not-existing-uid',
-			'expiredate' => '2018-06-28 10:13 UTC'
-			]);
+        $returnCode = $this->commandTester->execute([
```
PVince81: tabs

settermjd: Saw that. Took a little bit to change it.

phil-davis:
I have deja vu here  I am sure that I read code like this somewhere else in the last week while doing authentication related testing. Maybe it is in core somewhere also?

settermjd:
I created another PR around the same time as this one. I have a third but in a larger commit. I think that I should have done them all at the same time, in the same PR.

settermjd:
Thanks @phil-davis

Issue #39, opened 2018-07-11
Add options to the expire password command


phil-davis:
As suggested at https://github.com/owncloud/password_policy/issues/24#issuecomment-404195438

Add `--all` and `--group` options to the expire password ``occ`` command to make it more convenient to pre-expire passwords of whole groups of users or all existing users, specially when the ``password_policy`` app has been first enabled.

PVince81:
yes, makes sense.

There are already code examples in "occ files:scan", we should use the same format for the CLI arguments and same behavior in case a group doesn't exist.

Estimate: 0.5-1md (including tests, review, docs)

PR #77, opened 2018-07-31
Add options 'all' and 'group' to expire-password command

sharidas:
Adding options 'all' and 'group' to expire-password command.

### DIFF
--- a/lib/Command/ExpirePassword.php
+++ b/lib/Command/ExpirePassword.php
@@ -23,12 +23,15 @@
 namespace OCA\PasswordPolicy\Command;
 
 use OCP\IConfig;
+use OCP\IGroupManager;
+use OCP\IUser;
 use OCP\IUserManager;
 use OCA\PasswordPolicy\Db\OldPasswordMapper;
 use OCA\PasswordPolicy\Db\OldPassword;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use OCP\AppFramework\Utility\ITimeFactory;
 
@@ -52,27 +55,36 @@ class ExpirePassword extends Command {
 	/** @var \OCP\IUserManager */
 	protected $userManager;
 
+	/** @var IGroupManager */
+	private $groupManager;
+
 	/** @var ITimeFactory */
 	private $timeFactory;
 
 	/** @var OldPasswordMapper */
 	private $mapper;
 
+
 	/**
+	 * ExpirePassword constructor.
+	 *
 	 * @param IConfig $config
 	 * @param IUserManager $userManager
+	 * @param IGroupManager $groupManager
 	 * @param ITimeFactory $timeFactory
 	 * @param OldPasswordMapper $mapper
 	 */
 	public function __construct(
 		IConfig $config,
 		IUserManager $userManager,
+		IGroupManager $groupManager,
 		ITimeFactory $timeFactory,
 		OldPasswordMapper $mapper
 	) {
 		parent::__construct();
 		$this->config = $config;
 		$this->userManager = $userManager;
+		$this->groupManager = $groupManager;
 		$this->timeFactory = $timeFactory;
 		$this->mapper = $mapper;
 	}
@@ -82,18 +94,30 @@ class ExpirePassword extends Command {
 			->setName('user:expire-password')
 			->setDescription('Expire a user\'s password')
 			->addArgument(
-					'uid',
-					InputArgument::REQUIRED,
-					'The user\'s ownCloud uid (username).'
-				     )
-			->addArgument(
 					'expiredate',
 					InputArgument::OPTIONAL,
 					'The date and time when a password expires, e.g. "2019-01-01 14:00:00 CET".',
 					'-1 days'		// base.php sets timezone to utc, so
 										// make sure date is in the past
-				     )
-		;
+				)
+			->addOption(
+				'all',
+				null,
+				InputOption::VALUE_NONE,
+				'Will add password expiry to all known users. uid and group option will be discarded if all option is provided by user.'
+			)
+			->addOption(
+				'uid',
+				'u',
+				InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
+				'The user\'s uid is used. This option can be used as --uid "Alice" --uid "Bob"'
+			)
+			->addOption(
+				'group',
+				'g',
+				InputOption::VALUE_OPTIONAL | InputOption::VALUE_IS_ARRAY,
+				'Add password expiry to user(s) under group(s). This option can be used as --group "foo" --group "bar" to add expiry passwords for users in group foo and bar. If uid option (eg: --uid "user1") is passed with group, then uid will also be processed.'
+			);
 	}
 
 	/**
@@ -106,30 +130,84 @@ class ExpirePassword extends Command {
 	 * @throws \OCP\PreConditionNotMetException
 	 */
 	protected function execute(InputInterface $input, OutputInterface $output) {
-		$uid = $input->getArgument('uid');
-
-		/** @var $user \OCP\IUser */
-		$user = $this->userManager->get($uid);
-
-		if ($user === null) {
-			$output->writeln("<error>Unknown user: $uid</error>");
-			return self::EX_NOUSER;
-		}
-
-		if (!$user->canChangePassword()) {
-			$output->writeln("<error>The user's backend doesn't support password changes. The password cannot be expired for user: $uid</error>");
-			return self::EX_GENERAL_ERROR;
-		}
+		$groups = $input->getOption('group');
+		$allUsers = $input->getOption('all');
+		$uids = $input->getOption('uid');
 
 		$expireDate = new \DateTime();
-		$expireDate->setTimezone(new \DateTimeZone('UTC'));
-		$expireDate->setTimestamp($this->timeFactory->getTime());
-		$expireDate->modify($input->getArgument('expiredate'));
-
 		$oldDate = new \DateTime();
-		$oldDate->setTimezone(new \DateTimeZone('UTC'));
+		$timeZone = new \DateTimeZone('UTC');
+		$inputExpireDate = $input->getArgument('expiredate');
+
+		//This array will hold the uids of users process when group option is passed
+		$users = [];
+		if ($allUsers !== false) {
+			$this->userManager->callForAllUsers(function (IUser $user) use ($expireDate, $oldDate, $timeZone, $inputExpireDate, $output, &$users) {
+				if (isset($users[$user->getUID()]) === false) {
+					if ($user->canChangePassword()) {
+						$calculatedExpireDate = $this->setPasswordExpiry($expireDate, $oldDate, $timeZone, $inputExpireDate, $user);
+						$output->writeln('The password for ' . $user->getUID() . ' is set to expire on ' . $calculatedExpireDate->format('Y-m-d H:i:s T') . '.');
+						$users[$user->getUID()] = true;
+					}
+				}
+			});
+		} else {
+			if (\count($groups) >= 1) {
+				foreach ($groups as $group) {
+					if ($this->groupManager->groupExists($group) === true) {
+						foreach ($this->groupManager->findUsersInGroup($group) as $user) {
+							if (isset($users[$user->getUID()]) === false) {
+								if ($user->canChangePassword()) {
+									$calculatedExpireDate = $this->setPasswordExpiry($expireDate, $oldDate, $timeZone, $inputExpireDate, $user);
+									$output->writeln('The password for ' . $user->getUID() . ' is set to expire on ' . $calculatedExpireDate->format('Y-m-d H:i:s T') . '.');
+									$users[$user->getUID()] = true;
+								}
+							}
+						}
+					} else {
+						$output->writeln("Ignoring missing group $group");
+					}
+				}
+			}
+
+			if (\count($uids) > 0) {
+				foreach ($uids as $uid) {
+					/** @var $user \OCP\IUser */
+					$user = $this->userManager->get($uid);
+
+					if ($user === null) {
+						$output->writeln("<error>Unknown user: $uid</error>");
+					} else {
+						if (isset($users[$user->getUID()])) {
+							continue;
+						}
+						if (!$user->canChangePassword()) {
+							$output->writeln("<error>The user's backend doesn't support password changes. The password cannot be expired for user: $uid</error>");
+						} else {
+							$calculatedExpireDate = $this->setPasswordExpiry($expireDate, $oldDate, $timeZone, $inputExpireDate, $user);
+							$output->writeln('The password for ' . $user->getUID() . ' is set to expire on ' . $calculatedExpireDate->format('Y-m-d H:i:s T') . '.');
+						}
+					}
+				}
+			}
+		}
+	}
+
+	/**
+	 * @param \DateTime $expireDate
+	 * @param \DateTime $oldDate
+	 * @param \DateTimeZone $timeZone
+	 * @param $inputExpireDate
+	 * @param IUser $user
+	 */
+	protected function setPasswordExpiry(\DateTime $expireDate, \DateTime $oldDate, \DateTimeZone $timeZone, $inputExpireDate , IUser $user) {
+		$expireDate->setTimezone($timeZone);
+		$expireDate->setTimestamp($this->timeFactory->getTime());
+		$expireDate->modify($inputExpireDate);
+
+		$oldDate->setTimezone($timeZone);
 		$oldDate->setTimestamp($this->timeFactory->getTime());
-		$oldDate->modify($input->getArgument('expiredate'));
+		$oldDate->modify($inputExpireDate);
 
 		if ($this->config->getAppValue('password_policy', 'spv_user_password_expiration_checked', false) === 'on') {
 			$delta = $this->config->getAppValue('password_policy', 'spv_user_password_expiration_value', 90);
@@ -137,7 +215,7 @@ class ExpirePassword extends Command {
 		}
 
 		$this->config->deleteUserValue(
-			$uid,
+			$user->getUID(),
 			'password_policy',
 			'forcePasswordChange'
 		);
@@ -145,15 +223,11 @@ class ExpirePassword extends Command {
 		// add a dummy password in the user_password_history so the cron job
 		// can notify about the expiration of the password.
 		$oldPassword = new OldPassword();
-		$oldPassword->setUid($uid);
+		$oldPassword->setUid($user->getUID());
 		$oldPassword->setPassword(OldPassword::EXPIRED);
 		$oldPassword->setChangeTime($oldDate->getTimestamp());
 		$this->mapper->insert($oldPassword);
 
-		// show expire date if it was given
-		if ($input->hasArgument('expiredate')) {
-			$output->writeln("The password for $uid is set to expire on ". $expireDate->format('Y-m-d H:i:s T').'.');
-		}
-		return 0;
+		return $expireDate;
 	}
 }
--- a/tests/Command/ExpirePasswordTest.php
+++ b/tests/Command/ExpirePasswordTest.php
@@ -26,6 +26,7 @@ use OCA\PasswordPolicy\Command\ExpirePassword;
 use OCA\PasswordPolicy\Db\OldPasswordMapper;
 use OCA\PasswordPolicy\Db\OldPassword;
 use OCP\IConfig;
+use OCP\IGroupManager;
 use OCP\IUser;
 use OCP\IUserManager;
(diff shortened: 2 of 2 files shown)


lib/Command/ExpirePassword.php
```diff
@@ -83,17 +95,28 @@ protected function configure() {
...
+				'all',
+				null,
+				InputOption::VALUE_NONE,
+				'Will add password expiry to all known users.'
+			)
+			->addOption(
+				'groups',
+				'g',
+				InputArgument::OPTIONAL,
+				'Add password expiry to user(s) under group(s). This option can be used as --groups=foo,bar to scan groups foo and bar'
```
phil-davis: It needs to allow for comma in a group name. There is code in ``core`` repo ``core/Command/User/Add.php`` that has a ``group``  option that can be repeated on the command line.
```
--group foo --group bar
```

mmattel:
Please open a doc issue so that this extension gets documented when merged

lib/Command/ExpirePassword.php
sharidas: @phil-davis It does allow group names separated by comma. For example you can provide:
`./occ user:expire-password --groups=group1,group2    "2019-01-01 14:00:00 CET"`. Hope I understood your query correctly?

phil-davis: My comment is about a group that contains a comma - group called ``finance,Sydney``

sharidas: @phil-davis Ouch... Thanks for pointing that... :+1:

sharidas: Done. Also updated the test to have group name with comma.

lib/Command/ExpirePassword.php
```diff
@@ -106,54 +129,110 @@ protected function configure() {
...
-		}
-
-		if (!$user->canChangePassword()) {
-			$output->writeln("<error>The user's backend doesn't support password changes. The password cannot be expired for user: $uid</error>");
-			return self::EX_GENERAL_ERROR;
+		$groups = $input->getOption('groups');
+		$allUsers = $input->getOption('all');
+
+		if ($allUsers !== false) {
+			$users = $this->userManager->search('');
```
PVince81: can we use `callForAllUsers` instead ? the search will not scale for bigger setups

sharidas:
Fixes issue: https://github.com/owncloud/password_policy/issues/39

lib/Command/ExpirePassword.php
```diff
@@ -106,54 +129,110 @@ protected function configure() {
...
+	protected function getAllUsersFromGroup($group) {
+		$count = 0;
+		$users = [];
+		foreach ($this->groupManager->findUsersInGroup($group) as $user) {
+			\array_push($users, $user);
+			$count++;
+			//Take 200 users at a time
+			if ($count > 199) {
+				yield $users;
+				$count = 1;
```
PVince81: shouldn't this be 0 ? better write a unit test to confirm that this method works properly when crossing page borders

lib/Command/ExpirePassword.php
```diff
@@ -106,54 +129,110 @@ protected function configure() {
...
-		if ($input->hasArgument('expiredate')) {
-			$output->writeln("The password for $uid is set to expire on ". $expireDate->format('Y-m-d H:i:s T').'.');
+	/**
+	 * @param InputInterface $input
+	 * @param OutputInterface $output
+	 * @param IUser[] $users
+	 */
+	protected function setPasswordExpiry(InputInterface $input, OutputInterface $output, $users) {
+		foreach ($users as $user) {
+			$expireDate = new \DateTime();
```
PVince81: feels a bit overkill to reinitialize everything from the command in this method. this method has too many responsibilities. In my view, this method should only be responsible of setting the expiration for a list of users, not doing anything with $input or $output.

this this current code, you'd set the timestamp to a different time for every user as you iterate... I'd prefer consistency here

lib/Command/ExpirePassword.php
```diff
@@ -83,17 +95,28 @@ protected function configure() {
...
-		;
+				)
+			->addOption(
+				'all',
+				null,
+				InputOption::VALUE_NONE,
+				'Will add password expiry to all known users.'
+			)
+			->addOption(
+				'groups',
```
phil-davis: In ``core``  repo ``core/Command/User/Add.php`` the option is called just ``group`` because you specify just 1 group name after each ``--group`` : 
```
--group Finance,Sydney --group HR,Sydney --group Finance,Berlin
```
Should the option here be just ``group`` ?

tests/Command/ExpirePasswordTest.php
```diff
@@ -78,6 +82,104 @@ public function testExpirePasswordUserNotExisting() {
...
+		$this->commandTester->execute([
+			'--groups' => ['group1', 'group2', 'group4,foo', 'group3'],
+			'uid' => '',
+			'expiredate' => '2019-01-01 14:00:00 CET'
+		]);
+
+		$output = $this->commandTester->getDisplay();
+		$expectedOutput = "The password for user1 is set to expire on 2019-01-01 14:00:00 UTC.
+The password for user2 is set to expire on 2019-01-01 14:00:00 UTC.
+The password for user1 is set to expire on 2019-01-01 14:00:00 UTC.
```
phil-davis: I guess this behavior does not hurt - if the user is a member of multiple of the groups then they get their password expired multiple times. Probably there will be multiple similar entries in the password history for the user.

lib/Command/ExpirePassword.php
sharidas: I have updated the code. Wrote another method which would call `setPasswordExpiry`. The reinitialise of `expireDate`  and `oldDate` are now taken care once by the new function. Hence removed the usage of `input` and `output` from `setPasswordExpiry()` method.

lib/Command/ExpirePassword.php
sharidas: Renamed it to `group`.

tests/Command/ExpirePasswordTest.php
sharidas: Updated the change. I would appreciate if you can have a look.

jvillafanez:
We should include a `-u` option to provide a list of users the same way we can provide a list of groups, and leave the argument for compatiblity. It feels bad that you can provide several groups but not a couple of users.

We also need to clarify in the help the behaviour: what if `--all` and `-g` options are used? I expect that, if the `--all` option is used the `-g` (and `-u`?) options will be ignored

lib/Command/ExpirePassword.php
```diff
@@ -106,54 +129,128 @@ protected function configure() {
...
+		$groups = $input->getOption('group');
+		$allUsers = $input->getOption('all');
+
+		if ($allUsers !== false) {
+			$users = [];
+			$function = function (IUser $user) use (&$users) {
+				$users[] = $user;
+			};
+
+			$this->userManager->callForAllUsers($function);
```
jvillafanez: This is bad. You shouldn't use the callback to collect all the users. The purpose of using a callback is to avoid this precisely, to avoid having a list of all users in memory.
```
$this->userManager->callForAllUsers(function(IUser $user) use ($input, $output) {
  $this->setPasswordExpireForUser($input, $output, $user);
});
```
Note that the callback function process only one user.

lib/Command/ExpirePassword.php
```diff
@@ -106,54 +129,128 @@ protected function configure() {
...
+			$this->setPasswordExpiryForUsers($input, $output, $users);
+		} elseif (\count($groups) >= 1) {
+			$users = [];
+
+			foreach ($groups as $group) {
+				if ($this->groupManager->groupExists($group) === false) {
+					$output->writeln("Group name $group doesn't exist");
```
jvillafanez: this doesn't feel right... you provide A, B, C and D, but B doesn't exists. Why A is processed but C and D not?
The user isn't warned about this behaviour. I might expect that all the existing groups are processed, or it might be acceptable if the command aborts if one group is missing, but abort the command half way seems wrong.

lib/Command/ExpirePassword.php
```diff
@@ -106,54 +129,128 @@ protected function configure() {
...
+				$users[] = $user;
+			};
+
+			$this->userManager->callForAllUsers($function);
+			$this->setPasswordExpiryForUsers($input, $output, $users);
+		} elseif (\count($groups) >= 1) {
+			$users = [];
+
+			foreach ($groups as $group) {
+				if ($this->groupManager->groupExists($group) === false) {
```
jvillafanez: It's a bit better to check for `true` because we want to process the groups only if the group exists. If the function returns something else, we don't want to process the group.
(Just saying, as I expect some changes around and this one is easy. It isn't important because the method should only return true or false)

lib/Command/ExpirePassword.php
```diff
@@ -106,54 +129,128 @@ protected function configure() {
...
+			}
 		}
 
-		if (!$user->canChangePassword()) {
-			$output->writeln("<error>The user's backend doesn't support password changes. The password cannot be expired for user: $uid</error>");
-			return self::EX_GENERAL_ERROR;
+		return 0;
+	}
+
+	protected function getAllUsersFromGroup($group) {
```
jvillafanez: this is overcomplicating things. The purpose of the generator is to avoid creating a large array in order to consume less memory. If the array is already created, wrapping it with a generator is pointless, and it's better (and easier) to pass the array along.

lib/Command/ExpirePassword.php
```diff
@@ -106,54 +129,128 @@ protected function configure() {
...
+			yield $users;
 		}
+	}
 
+	/**
+	 * @param InputInterface $input
+	 * @param OutputInterface $output
+	 * @param IUser[] $users
+	 */
+	private function setPasswordExpiryForUsers(InputInterface $input, OutputInterface $output, $users) {
```
jvillafanez: I'd prefer to remove the `InputInterface` parameter and use whatever the function needs. The function doesn't need access to the whole object.
Something similar for the `OutputInterface` although in this case the function might really need to write something, so it's probable easier to use the `OutputInterface`.

lib/Command/ExpirePassword.php
```diff
@@ -106,54 +129,128 @@ protected function configure() {
...
 		$expireDate = new \DateTime();
-		$expireDate->setTimezone(new \DateTimeZone('UTC'));
+		$oldDate = new \DateTime();
+		$timeZone = new \DateTimeZone('UTC');
+		$inputExpireDate = $input->getArgument('expiredate');
+		foreach ($users as $user) {
+			$calculatedExpireDate = $this->setPasswordExpiry($expireDate, $oldDate, $timeZone, $inputExpireDate, $user);
+			if ($input->hasArgument('expiredate')) {
```
jvillafanez: Better add a comment here. It's weird that you get an argument, use it and later check if the argument is present.
I guess this `hasArgument` checks if the argument has been explicitly provided, while the `getArgument` might return the default value. I'm not sure if we should just write the message regardless of this...

lib/Command/ExpirePassword.php
```diff
@@ -106,54 +129,128 @@ protected function configure() {
...
+	}
+
+	/**
+	 * @param \DateTime $expireDate
+	 * @param \DateTime $oldDate
+	 * @param \DateTimeZone $timeZone
+	 * @param $inputExpireDate
+	 * @param IUser $user
+	 */
+	protected function setPasswordExpiry($expireDate, $oldDate, $timeZone, $inputExpireDate ,$user) {
```
jvillafanez: include datatypes in the parameters

tests/Command/ExpirePasswordTest.php
jvillafanez: I think we should fix this. From a UX perspective it feels wrong that the password is reset several times for the same user, specially if the command is run once. Technically, I'm not sure if  we're generating and storing useless data in the DB. In both cases people will think that it's a bug.

lib/Command/ExpirePassword.php
sharidas: Removed the usage of checking it. In the latest version of the PR, its done in the `execute`.

lib/Command/ExpirePassword.php
sharidas: Removed `setPasswordExpiryForUsers`. And reusing `setPasswordExpiry`, where only required parameters are used.

lib/Command/ExpirePassword.php
sharidas: Removed the usage of `getAllUsersFromGroup`. Iterating the output of `findUsersInGroup($group)`.

lib/Command/ExpirePassword.php
sharidas: In the latest version I removed the check to false. And used `true`. Also made sure that if the group doesn't exist, it wouldn't return from the `execute` method. The list of groups will be parsed and processed.

lib/Command/ExpirePassword.php
sharidas: The latest version of PR takes care of this issue.

lib/Command/ExpirePassword.php
sharidas: Updated `callForUsers` as per the comment.

sharidas:
Updated the PR to pass multiple `uid` so that multiple users can also be processed apart from `all` or `group`.

lib/Command/ExpirePassword.php
```diff
@@ -106,54 +130,96 @@ protected function configure() {
...
+		} elseif (\count($groups) >= 1) {
+			$users = [];
+
+			foreach ($groups as $group) {
+				if ($this->groupManager->groupExists($group) === true) {
+					foreach ($this->groupManager->findUsersInGroup($group) as $user) {
+						if (!\in_array($user->getUID(), $users)) {
+							$calculatedExpireDate = $this->setPasswordExpiry($expireDate, $oldDate, $timeZone, $inputExpireDate, $user);
+							$output->writeln('The password for ' . $user->getUID() . ' is set to expire on ' . $calculatedExpireDate->format('Y-m-d H:i:s T') . '.');
+							$users[] = $user->getUID();
```
jvillafanez: It's better to store the user ids as the key. We expect the user to not be in the array, let's say the 90% of the times (having the user in several of these groups shouldn't happen too often), so most of the times you'll need to traverse the whole array over and over (80-90% of the times), and the array will increase.

Storing the uids as keys will allow us to check the existence of the key via `isset($users[$uid])`, which will perform the check faster and in constant time regardless of the size of the array. You might want to store it like `$users[$user->getUID()] = true` for convenience. I don't think we'll need the user object later.

lib/Command/ExpirePassword.php
```diff
@@ -106,54 +130,96 @@ protected function configure() {
...
+		$oldDate = new \DateTime();
+		$timeZone = new \DateTimeZone('UTC');
+		$inputExpireDate = $input->getArgument('expiredate');
+
+		if ($allUsers !== false) {
+			$this->userManager->callForAllUsers(function (IUser $user) use ($expireDate, $oldDate, $timeZone, $inputExpireDate, $output) {
+				$calculatedExpireDate = $this->setPasswordExpiry($expireDate, $oldDate, $timeZone, $inputExpireDate, $user);
+				$output->writeln('The password for ' . $user->getUID() . ' is set to expire on ' . $calculatedExpireDate->format('Y-m-d H:i:s T') . '.');
+			});
+		} elseif (\count($groups) >= 1) {
```
jvillafanez: It might be useful to expire passwords for groups and users at the same time. In addition, the help doesn't say anything that suggest that this can't be done

lib/Command/ExpirePassword.php
```diff
@@ -106,54 +130,96 @@ protected function configure() {
...
+				if ($this->groupManager->groupExists($group) === true) {
+					foreach ($this->groupManager->findUsersInGroup($group) as $user) {
+						if (!\in_array($user->getUID(), $users)) {
+							$calculatedExpireDate = $this->setPasswordExpiry($expireDate, $oldDate, $timeZone, $inputExpireDate, $user);
+							$output->writeln('The password for ' . $user->getUID() . ' is set to expire on ' . $calculatedExpireDate->format('Y-m-d H:i:s T') . '.');
+							$users[] = $user->getUID();
+						}
+					}
+				} else {
+					$output->writeln("<error>Group name $group doesn't exist</error>");
```
jvillafanez: Maybe changing the message for "Ignoring missing group $group" sounds more like a warning instead of an error. I'm not sure if it's possible to change the "error" tag to a "warning" tag.

lib/Command/ExpirePassword.php
```diff
@@ -106,54 +130,96 @@ protected function configure() {
...
+			foreach ($uids as $uid) {
+				/** @var $user \OCP\IUser */
+				$user = $this->userManager->get($uid);
+
+				if ($user === null) {
+					$output->writeln("<error>Unknown user: $uid</error>");
+					return self::EX_NOUSER;
+				}
+
+				if (!$user->canChangePassword()) {
```
jvillafanez: This check should be shared among the groups and "all users". I mean, the behaviour must be same for the same user regardless of trying to change his password using the `-u` option or `-g` option (if he's member of the group) or `--all` option

lib/Command/ExpirePassword.php
```diff
@@ -106,54 +130,96 @@ protected function configure() {
...
+				}
+			}
+		} else {
+			foreach ($uids as $uid) {
+				/** @var $user \OCP\IUser */
+				$user = $this->userManager->get($uid);
+
+				if ($user === null) {
+					$output->writeln("<error>Unknown user: $uid</error>");
+					return self::EX_NOUSER;
```
jvillafanez: These return statements should be removed

jvillafanez:
We'll also need to review the wording for the help and errors / warnings.
If it isn't an error we should make sure that reading the message doesn't sound like an error.

lib/Command/ExpirePassword.php
sharidas: Didn't find a "warning" tag. These are the default ones available https://symfony.com/doc/current/console/coloring.html#using-color-styles. We can customise the colour. Like setting foreground and background colours.

sharidas:
Users can also pass `-all` and `uid` together. The current version of the PR will not allow to process users again for the `uid` user(s).

mmattel:
@sharidas in easy words, ``--all`` takes precedence over ``uid``.
If this is true, we have to put a note in the documentation to make it clear (like we have done it eg for files:scan).

sharidas:
@mmattel 
> in easy words, --all takes precedence over uid

Yes.

lib/Command/ExpirePassword.php
```diff
@@ -106,54 +130,105 @@ protected function configure() {
...
+				} else {
+					$output->writeln("Ignoring missing group $group");
+				}
+			}
 		}
-
-		if (!$user->canChangePassword()) {
-			$output->writeln("<error>The user's backend doesn't support password changes. The password cannot be expired for user: $uid</error>");
-			return self::EX_GENERAL_ERROR;
+		if (\count($uids) > 0) {
```
jvillafanez: If both the `--all` and `-u` options are provided, we should ignore the `-u` options, which means this code shouldn't be executed in that scenario.

lib/Command/ExpirePassword.php
```diff
@@ -106,54 +130,105 @@ protected function configure() {
...
+			foreach ($uids as $uid) {
+				/** @var $user \OCP\IUser */
+				$user = $this->userManager->get($uid);
+
+				//If it is processed by the group option we can skip this user
+				if (($user !== null) && isset($users[$user->getUID()])) {
```
jvillafanez: we should reorganize these conditions:
```
if ($user === null) {
....
} else {
  if (isset($users[$user->getUID()])) {
    // ignored user that has been processed already
    continue;
  }
  if (!$user->canChangePassword()) {
    // show error
  } else {
    // process user
  }
}
```
This should be easier to understand, and also we make sure all the possible conditions are properly covered

jvillafanez:
Question: will the expire date be the same for all the users?
The "The password for user1 is set to expire on 2019-01-01 14:00:00 UTC" message might be annoying if you have 100 users and you only change the userid.

If we don't aim to have a parseable output maybe something like "the following users will have their password expired on 2019-01-01 ....." and then provide the list of users would be better. Maybe @pmaier1 can give better feedback.

lib/Command/ExpirePassword.php
sharidas: Done

jvillafanez:
```
if ($allUser !== false) {
  ...
} else {
  if (\count($groups) > 0) {
    ...
  }

  if (\count($users) > 0) {
    ...
  }
}
```
This structure should be more consistent with what we want to do.

sharidas:
> This structure should be more consistent with what we want to do.

Done

PVince81:
This will need an entry in the release notes because now the occ command signature changed.

I had `occ user:expire-password user1` in my shell history and now this command doesn't output anything, not even an error message. I was very confused until I noticed that now I need to use `-u`.

@pmaier1 ^

PVince81:
I've raised https://github.com/owncloud/password_policy/issues/100 to make the command output something when args are missing instead of doing nothing

pmaier1:
> This will need an entry in the release notes because now the occ command signature changed.

Noted.

Issue #37, opened 2018-07-11
Symfony events for expiry and reminder


PVince81:
Trigger symfony events whenever an expiry notification is sent and also whenever the password expires.

PR #79, opened 2018-08-02
Add symfony events for about to expire and expired

sharidas:
Add new symfony events for passwords about to
expire and expired.

### DIFF
--- a/lib/Jobs/PasswordExpirationNotifierJob.php
+++ b/lib/Jobs/PasswordExpirationNotifierJob.php
@@ -31,6 +31,8 @@ use OCP\AppFramework\Utility\ITimeFactory;
 use OCA\PasswordPolicy\Db\OldPasswordMapper;
 use OCA\PasswordPolicy\Db\OldPassword;
 use OCA\PasswordPolicy\UserNotificationConfigHandler;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+use Symfony\Component\EventDispatcher\GenericEvent;
 
 class PasswordExpirationNotifierJob extends TimedJob {
 	/** @var OldPasswordMapper */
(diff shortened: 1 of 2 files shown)


sharidas:
This PR fixes -> https://github.com/owncloud/password_policy/issues/37

jvillafanez:
🚫 
The `Notifier` class isn't the right place to trigger the events. You'll be triggering the event each time the notification is rendered (web UI and API calls).

Note that we'll likely want information about the password (not really sure what kind of information though), but not about the notification.

You aren't triggering a "user.passwordAboutToExpire" event but a "user.notificationForPasswordAboutToExpireRendered"

sharidas:
Made the modifications in `HooksHandler`. I guess this is the place where the symfony events should be placed.

jvillafanez:
Not there neither. You'll be triggering the event when a password has changed, but I can change the passwords once a day and it isn't near its expiration time.

I'd check around the background job, unless there is a better place.

sharidas:
@jvillafanez I have updated the code by making change in background job.

jvillafanez:
Moving the event dispatch after the job has marked that the notification is sent feels a bit better. We also might need to send some information about the expired (or about to expire) password.

I'm not sure if we should add something in the event name to indicate that the event will be triggered by the brackground job, which means that it won't be immediate. The event might be sent way after the actual expiration date.

sharidas:
@jvillafanez 

> Moving the event dispatch after the job has marked that the notification is sent feels a bit better. We also might need to send some information about the expired (or about to expire) password.

- I have moved the change after the job has marked the notification is sent.
- Other than `expirationTime` I couldn't find relevant data related to password, to be passed in the event.

jvillafanez:
> Other than expirationTime I couldn't find relevant data related to password, to be passed in the event.

Well, my original candidate was the `OldPassword` object, but on a second thought I'm not sure if we want to expose this information to other apps since we want to avoid inter-app dependencies. Just remove the it from the event for now, we can add it later if needed.

Pinging @PVince81 for a behaviour review, mainly to confirm we want to trigger the events there (https://github.com/owncloud/password_policy/pull/79#issuecomment-411380585).

Issue #72, opened 2018-07-20
Improve config UI


pmaier1:
![password-policy-app](https://user-images.githubusercontent.com/21193500/43003398-b1b78a3c-8c2c-11e8-8fc5-b3319f09776a.png)

The password history part ("X different than last passwords") is
a) absolutely not understandable and
b) located wrong as this section is for both, user passwords and public link passwords.

We should find a more comprehensive wording and move that one down to "User password policies".

@PVince81

PVince81:
Estimate: 0.25md

PVince81:
Rename -> "Restrict to these characters"

PR #80, opened 2018-08-02
Improve config messages in settings

sharidas:
This change brings improvisation to the messages
in the password_policy setttings.
1. Moved "different than last passwords" under "User password policies"
2. Rephrased "different than last passwords" to "last passwords should not be used"
3. Rephrased "Define special characters" to "Restrict to these special characters"


sharidas:
Fixes issue -> https://github.com/owncloud/password_policy/issues/72

Issue #65, opened 2018-07-17
Password Policy Warnings at 1st login should change when password is wrong


davitol:
### Steps to reproduce

1. Login as admin
2. Set Password policy 3 different than last passwords and Force users to change their password on first login
3. Create a new user with pwd `a` and log in
4. Login page shows your password has expired and ask for a new pwd
5. Write down `a` as pwd in the 3 fields and click on save button
6. Warning says `Password must be different than the last passwords` (right behavior)
7. Write down `b` as pwd in the 3 fields and click on save button

### Expected behavior

 Warning should say `Old Password is wrong`

### Current behavior

 Warning says `Password must be different than the last passwords`

![screen shot 2018-07-17 at 16 49 53](https://user-images.githubusercontent.com/10596560/42825622-af016ed0-89e2-11e8-8ce2-bd4742ea2ff5.png)
![screen shot 2018-07-17 at 16 50 31](https://user-images.githubusercontent.com/10596560/42825636-b55e25f2-89e2-11e8-80fe-d86585b61f6e.png)
**After step 5 it always appears these warnings**
![screen shot 2018-07-17 at 16 50 46](https://user-images.githubusercontent.com/10596560/42825648-c00547ba-89e2-11e8-89c6-c190eed31de9.png)

PVince81:
This state could likely be detected by seeing that there are zero passwords in the history for this users.
We could also try using the "oc_accounts.last_login" value, unless it's already set at this state.

Setting to p4 as it's cosmetic, unless @pmaier1 thinks it's more critical

PVince81:
another thing to fix as part of this ticket: https://github.com/owncloud/password_policy/issues/50 (closed)

Basically: adjust messages related to first password login:
- [ ] "Old password is wrong"
- [ ] Replace "Your password expired" with "Please set a new password)

PVince81:
Estimate: 0.5md

PR #81, opened 2018-08-03
Update the messages for first time login

sharidas:
When first time login password change is enforced
and user tries to provide wrong passwords to the
current password, the message shown is
"Old password is wrong."

The first time login password reset is also improved to
"Please set a new password"

### DIFF
--- a/lib/Controller/PasswordController.php
+++ b/lib/Controller/PasswordController.php
@@ -93,6 +93,13 @@ class PasswordController extends Controller implements IAccountModuleController 
 			'redirect_url' => $redirect_url,
 			'webroot' => \OC::$WEBROOT
 		];
+
+		$user = $this->userSession->getUser();
+		if ($user !== null) {
+			if ($this->config->getUserValue($user->getUID(), 'password_policy', 'firstLoginPasswordChange') === '1') {
+				$params['firstLogin'] = true;
+			}
+		}
 		if (empty($params['redirect_url'])) {
 			$params['redirect_url'] = $this->urlGenerator->getAbsoluteURL('');
 		}
@@ -138,6 +145,14 @@ class PasswordController extends Controller implements IAccountModuleController 
 	 * @return TemplateResponse|RedirectResponse
 	 */
 	public function update($current_password, $new_password, $confirm_password, $redirect_url) {
+		$user = $this->userSession->getUser();
+		if(($user !== null) && !$this->userManager->checkPassword($user->getUID(), $current_password)) {
+			return $this->createPasswordTemplateResponse(
+				$redirect_url,
+				$this->l10n->t('Old password is wrong.')
+			);
+		}
+
 		if ($new_password !== $confirm_password) {
 			return $this->createPasswordTemplateResponse(
 				$redirect_url,
@@ -149,15 +164,6 @@ class PasswordController extends Controller implements IAccountModuleController 
 			return $this->createPasswordTemplateResponse(
 				$redirect_url,
 				$this->l10n->t('Password must be different than the old password.')
-			);
-		}
-
-		$user = $this->userSession->getUser();
-
-		if(!$this->userManager->checkPassword($user->getUID(), $current_password)) {
-			return $this->createPasswordTemplateResponse(
-				$redirect_url,
-				$this->l10n->t('The current password is incorrect.')
 			);
 		}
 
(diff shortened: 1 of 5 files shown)


sharidas:
Fixes issue -> https://github.com/owncloud/password_policy/issues/65

templates/password.php
```diff
@@ -33,8 +33,7 @@
 <form id="password_policy" method="post">
 	<fieldset>
 		<h1 class="warning">
-			<div><?php p($l->t('Your password has expired.')); ?></div>
```
PVince81: we need to keep this message when the password has actually expired

it seems this PR completely removes the existing messages ?

sharidas:
Screenshot of the change 
firstlogin_setpasswd

lib/HooksHandler.php
```diff
@@ -257,7 +257,10 @@ public function savePasswordForCreatedUser(GenericEvent $event) {
 	/**
 	 * Flags the session to require a password change
 	 */
-	protected function forcePasswordChange() {
+	protected function forcePasswordChange($firstLogin = null, IUser $user = null) {
+		if (($firstLogin === true) && ($user !== null)) {
+			$this->config->setUserValue($user->getUID(), 'password_policy', 'firstLoginPasswordChange', '1');
```
PVince81: you likely don't need this and check the existing value of "last_login" in the user object / accounts table

PVince81: if last_login is null it means the user has never logged in before

sharidas: Initially I thought to use the same. But then I believe when user tries to login with username + password from the webUI,  the code reaches
- https://github.com/owncloud/core/blob/master/lib/private/User/Session.php#L520
- https://github.com/owncloud/core/blob/master/lib/private/User/Manager.php#L254
And hence the "last_login" will be updated. So if we try to use "last_login" the value will be a positive integer. ( Cross verified just now ).

PVince81: Do you mean that even with failed login we do set the "last_login" value ? That doesn't sound right as it would likely also trigger skeleton creation.

Mind debugging to confirm, in case there's another hidden bug ?

sharidas: Sure I can debug this. But my comment above ( https://github.com/owncloud/password_policy/pull/81#discussion_r207847148) wasn't with respect to failed login. The scenario I had was like this:
- enforced first login password change in password policy settings
- Create a user `user1`
- Try to login from web ui as `user1`.
- This would populate "last_login" by the time code reaches https://github.com/owncloud/password_policy/blob/master/lib/HooksHandler.php#L317

sharidas: The skeleton is copied by the time password_policy code reaches https://github.com/owncloud/password_policy/blob/master/lib/HooksHandler.php#L317

sharidas: I think the behaviour with the "last_login" is ok with respect to my comments. Because the "last_login" will be updated https://github.com/owncloud/core/blob/master/lib/private/User/Session.php#L529. Since its the first time login and its validated, the code reaches https://github.com/owncloud/core/blob/master/lib/private/User/Session.php#L532 followed by https://github.com/owncloud/core/blob/master/lib/private/User/Session.php#L465. By this time the skeleton is already copied. 

Now password_policy triggers the password reset template. So "last_login" value is already updated for the user.

So I believe the code behaviour is correct.

PVince81: Okay, thanks for the details.

In that case the proposed approach makes sense.

lib/Controller/PasswordController.php
```diff
@@ -93,6 +93,13 @@ protected function createPasswordTemplateResponse($redirect_url, $error = null)
 			'redirect_url' => $redirect_url,
 			'webroot' => \OC::$WEBROOT
 		];
+
+		$user = $this->userSession->getUser();
+		if ($user !== null) {
+			if ($this->config->getUserValue($user->getUID(), 'password_policy', 'firstLoginPasswordChange') === '1') {
```
PVince81: another approach would be the reverse: set a flag after the first login password was set. but ok, this works too

PVince81:
@sharidas please adjust unit tests

sharidas:
Except for the template -> templates/password.php, rest of the php files are covered in the unit tests. I believe this PR does cover rest of the PHP files in the code.

Issue #9, opened 2018-01-24
strip_tags when saving special characters value disallows the use of < and > in passwords.


mdusher:
strip_tags() is used when saving the special characters definition on this line:
https://github.com/owncloud/password_policy/blob/master/lib/Controller/SettingsController.php#L73

This modifies the value the user has entered and can strip more characters than just `<>` depending on where they are entered.

**Problem**
If a user would like to use `<` and `>` as special characters in their password policy they can't.

**What happens**
If `<` and `>` are added to the defined special characters, they are stripped from the list (and any characters following them in the list).

**How to reproduce**
1. Type `#><!@` or `#<!>@` into the "Define special characters" field.
2. Click "Save".
3. Reload the page and the characters after `<` will disappear

pako81:
Confirmed. Tested with 10.0.4 + Password Policy 1.1.0.

cdamken:
@PVince81 Could you please assign an engineer to this?

phil-davis:
I For me, the ``>`` is working OK as a required special character in a password.
It is only when you put ``<`` in the special characters list that there is the trouble described.

PVince81:
Estimated: 0.25-0.5md

sharidas:
Removal of `strip_tags`, does resolve the issue.

PR #82, opened 2018-08-03
Remove strip_tags to include tags as special chars

sharidas:
If user wants to add php/html tags as special chars,
then strip_tags should not be used. This patch
addresses the issue.


sharidas:
Fixes issue -> https://github.com/owncloud/password_policy/issues/9

Issue #69, opened 2018-07-18
Clean up after user deletion


PVince81:
- ~~delete notifications~~: https://github.com/owncloud/password_policy/issues/53 => https://github.com/owncloud/core/issues/32243
- [ ] delete old passwords: https://github.com/owncloud/password_policy/issues/28

Estimate: ~~0.5md for both together~~ 0.25md

sharidas:
A clean solution for clearing notification table would be to have a method in NotificationManager which would remove the entries of user from notification table. A ticket opened at https://github.com/owncloud/core/issues/32243.

PR #83, opened 2018-08-07
Clean up tables after user deletion

sharidas:
Clean up the user entries in the table after
the user is deleted.

### DIFF
--- a/lib/HooksHandler.php
+++ b/lib/HooksHandler.php
@@ -328,4 +328,17 @@ class HooksHandler {
 			$this->forcePasswordChange(true, $user);
 		}
 	}
+
+	/**
+	 * This method removes the entries which remain in oc_user_password_history
+	 * table and oc_notifications table after the deletion of the user
+	 *
+	 * @param GenericEvent $event
+	 */
+	public function removeUserEntriesFromTable(GenericEvent $event) {
+		$this->fixDI();
+		$uid = $event->getArgument('uid');
+
+		$this->oldPasswordMapper->cleanUserHistory($uid);
+	}
 }
--- a/tests/Db/OldPasswordMapperTest.php
+++ b/tests/Db/OldPasswordMapperTest.php
@@ -25,6 +25,8 @@ namespace OCA\PasswordPolicy\Tests\Db;
 
 use OCA\PasswordPolicy\Db\OldPassword;
 use OCA\PasswordPolicy\Db\OldPasswordMapper;
+use OCP\DB\QueryBuilder\IExpressionBuilder;
+use OCP\DB\QueryBuilder\IParameter;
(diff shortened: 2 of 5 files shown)


sharidas:
Fixes -> https://github.com/owncloud/password_policy/issues/69

lib/Db/ClearUserPasswordHistory.php
```diff
@@ -0,0 +1,45 @@
...
+ *
+ */
+
+namespace OCA\PasswordPolicy\Db;
+
+use OCP\AppFramework\Db\Mapper;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\IDBConnection;
+
+class ClearUserPasswordHistory extends Mapper {
```
PVince81: no need for a new mapper, just add a method to the existing one

sharidas: Moved this method to existing mapper in the app.

tests/Db/OldPasswordMapperTest.php
```diff
@@ -207,4 +209,31 @@ public function testGetPasswordsAboutToExpireAllMarked() {
 		$this->assertSame($uid, $latestPassword->getUid());
 		$this->assertLessThan($baseTime + 130, $latestPassword->getChangeTime());
 	}
+
+	public function testCleanUserHistory() {
```
jvillafanez: I don't think this test does what we want it to do.
The method should delete specific DB entries; what you're testing here is that the query builder is called in a predefined way. It seems 2 different things to me.

If I remove the `execute` statement, the test won't fail, but the method won't do what is supposed to do.
If I change the implementation to use a raw SQL statement (because random reasons), even if it works properly, the test will fail. Why the test should complain when the user entries are deleted?

lib/HooksHandler.php
```diff
@@ -117,6 +117,7 @@ private function fixDI() {
 			$this->session = \OC::$server->getSession();
 			$this->notificationManager = \OC::$server->getNotificationManager();
 			$this->userNotificationConfigHandler = new UserNotificationConfigHandler($this->config);
+			$this->connection = \OC::$server->getDatabaseConnection();
```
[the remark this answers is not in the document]
sharidas: Removed.

tests/Db/OldPasswordMapperTest.php
sharidas: Modified the test. Now the test has 3 users. And removed one of them. Verified the removal and existence of other users.

PVince81:
conflicts

jvillafanez:
I assume this has been tested and confirmed it works properly.

jvillafanez:
👍

PR #92, opened 2018-08-22
Remember errors in lint of PHP files

phil-davis:
The bash ``find`` command ignores the individual exit status of the ``exec``ed command that is done each time in the loop. So we are not getting a Travis fail when there is a lint problem found in the PHP code.

Taken from ideas here: https://unix.stackexchange.com/questions/195677/bash-error-handling-on-find-exec

When something goes wrong with ``php -l`` then put the related file names into a file. At the end, if the file exists, write out the file names with problems and exit with an error status.

### DIFF
--- a/lib/Controller/SettingsController.php
+++ b/lib/Controller/SettingsController.php
@@ -85,7 +85,9 @@ class SettingsController extends Controller implements ISettings {
 			if ($this->request->getParam($key) !== null) {
 				if ($key !== 'spv_def_special_chars_value' && \substr($key, -6) === '_value') {
 					$value = \min(\max(0, (int)$this->request->getParam($key)), 255);
-					if (isset(self::CONVERSIONS[$key]['in'])) {
+					if (array_key_exists($key, self::CONVERSIONS)
+						&& array_key_exists('in', self::CONVERSIONS[$key])
+					) {
 						$convertFuncName = self::CONVERSIONS[$key]['in'];
 						$value = $this->$convertFuncName($value);
 					}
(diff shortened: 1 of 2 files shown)


phil-davis:
This now makes the Travis job fail because there is an error reported for PHP 5.6 in ``SettingsController.php``
https://travis-ci.org/owncloud/password_policy/jobs/419086783
```
PHP Fatal error:  Cannot use isset() on the result of an expression (you can use "null !== expression" instead) in ./lib/Controller/SettingsController.php on line 88
Fatal error: Cannot use isset() on the result of an expression (you can use "null !== expression" instead) in ./lib/Controller/SettingsController.php on line 88
Errors parsing ./lib/Controller/SettingsController.php
```

Indeed, if you run the following code on PHP 5.6 at http://phptester.net/ :
```
<?php
class MyClass {
	const CONVERSIONS = [
		'xyz' => [
			'in' => 'daysToSeconds',
			'out' => 'secondsToDays',
		],
	];

	public function isItSet($key) {
		if (isset(self::CONVERSIONS[$key]['out'])) {
			echo "it is set";
		} else {
			echo "nope";
		}
	}
}

$a = new MyClass;
$a->isItSet('xyz');
```
you get:
```
FATAL ERROR Cannot use isset() on the result of an expression (you can use "null !== expression" instead) on line number 11
```

This is an existing issue in the code that will be a problem if it is run on PHP 5.6

phil-davis:
Using ``!== null`` results in PHP notices about keys that do not exist. Using ``array_key_exists`` at each level of the array works on PHP 5.6 and 7.*

phil-davis:
This is ready to review. I want to make sure the existing CI test is working before I port stuff to drone (and do it in a different way...)

phil-davis:
Settings changes have been manually tested in the webUI, including changing the days until password expires values. The values are saved OK.
@PVince81 as discussed.

PR #98, opened 2018-08-23
PHP style and PHP unit on drone

phil-davis:
Add ``.drone.yml`` that runs php-cs-fixer ("equivalent/better than" lint on Travis) and PHP unit tests (on sqLite, mySQL and pgSQL). See issue #101 for the reason that Oracle is commented out.

codecov is commented out for now. It can be added by whoever manages the "secrets".

I suggest we:
- commit this
- change the GitHub CI so drone is required and Travis is optional
- then I will make a PR to remove the Travis stuff

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,209 @@
+workspace:
+  base: /var/www/owncloud
+  path: apps/password_policy
+
+branches: [master, release*, release/*]
+
+pipeline:
+  install-server:
+    image: owncloudci/core
+    version: ${OC_VERSION}
+    pull: true
+    db_type: ${DB_TYPE}
+    db_name: ${DB_NAME}
+    db_host: ${DB_TYPE}
+    db_username: autotest
+    db_password: owncloud
+    exclude: apps/password_policy
+    when:
+      matrix:
+        NEED_SERVER: true
+
+  install-app:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    commands:
+      - cd /var/www/owncloud/
+      - php occ a:l
+      - php occ a:e password_policy
+      - php occ a:e testing
+      - php occ a:l
+      - php occ config:system:set trusted_domains 1 --value=owncloud
+      - php occ log:manage --level 0
+    when:
+      matrix:
+        NEED_INSTALL_APP: true
+
+  fix-permissions:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    commands:
+    - chown www-data /var/www/owncloud -R
+
+  owncloud-log:
+    image: owncloud/ubuntu:16.04
+    detach: true
+    pull: true
+    commands:
+      - tail -f /var/www/owncloud/data/owncloud.log
+    when:
+      matrix:
+        NEED_SERVER: true
+
+  php-cs-fixer:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    commands:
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -0,0 +1,227 @@
...
+  install-server:
+    image: owncloudci/core
+    version: ${OC_VERSION}
+    pull: true
+    db_type: ${DB_TYPE}
+    db_name: ${DB_NAME}
+    db_host: ${DB_TYPE}
+    db_username: autotest
+    db_password: owncloud
+    exclude: apps/password_policy
```
patrickjahns: Why the exclude? this is not required for general apps ( only the ones that are shipped in the qa-tarball which is market / notifications / encryption ). It doesn't hurt - but is not required and might confuse people when copy+pasting it around

.drone.yml
```diff
@@ -0,0 +1,227 @@
...
+      matrix:
+        TEST_SUITE: php-cs-fixer
+
+  phpunit-tests:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    environment:
+      - PHP_VERSION=${PHP_VERSION}
+    commands:
+      - cd /var/www/owncloud/apps/password_policy
```
patrickjahns: not needed - default directory workspace will be `/var/www/owncloud/apps/password_policy` as it is defined in the yml

.drone.yml
```diff
@@ -0,0 +1,227 @@
...
+      - tail -f /var/www/owncloud/data/owncloud.log
+    when:
+      matrix:
+        NEED_SERVER: true
+
+  php-cs-fixer:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    commands:
+      - cd /var/www/owncloud/apps/password_policy
```
patrickjahns: not needed - default directory workspace will be `/var/www/owncloud/apps/password_policy` as it is defined in the yml

.drone.yml
```diff
@@ -0,0 +1,227 @@
...
+      - php occ log:manage --level 0
+    when:
+      matrix:
+        NEED_INSTALL_APP: true
+
+  fix-permissions:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    commands:
+    - chown www-data /var/www/owncloud -R
```
patrickjahns: WHY? we do not have any job that requires `www-data` as user

individual-it: not yet, acceptance tests run as `www-data`

.drone.yml
phil-davis: I copy-pasted it from somewhere! So exactly this problem, I don't really know if it is needed.

.drone.yml
phil-davis: I wonder why we need to do that, in general. The acceptance tests should be hitting APIs and logging in to the webUI. So they should run as "any OS client user" that gives credentials to the API or logs in to the webUI.
@individual-it why is it so?

phil-davis: Running the acceptance test script works fine without it being ``www-data`` - as it should.
The only trick sometimes will be when ``chown www-data /var/www/owncloud -R`` has happened. If that happens to take read access away for ``/var/www/owncloud/tests/acceptance`` then there would be a problem. That would be just because the acceptance test scripts are "hiding" inside the server code. Some day that may get separated and thus avoid even this possible issue.

See PR #112 that removes the excess ``.drone.yml`` code.

PR #109, opened 2018-08-27
Delete all app config settings in 1 call to the testing app

phil-davis:
using new method from core.


phil-davis:
Note: CI will need to be run again when ``AppConfigHelper::deleteAppConfigs()`` comes in the core QA tarball.

phil-davis:
Note: method ``deletePasswordPolicySetting()`` has been left in the code, because there will be API acceptance tests coming that will likely want to make sure that policy settings can be set/cleared using the ``occ`` command. So this method will be used...

PR #111, opened 2018-08-27
Test steps to check underlying policy settings

phil-davis:
to make sure it is not just the pretty webUI ticks, but that app settings are actually happening.


individual-it:
Looks good, should have it in future in core

Issue #101, opened 2018-08-24
OldPasswordMapperTest failing on Oracle


phil-davis:
When I run these unit tests on Oracle, it gives errors as in this drone job
https://drone.owncloud.com/owncloud/password_policy/14/37

```
php -d zend.enable_gc=0  vendor/bin/phpunit --configuration ./tests/unit/phpunit.xml --testsuite unit
PHPUnit 5.7.27 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.1.18-1+ubuntu16.04.1+deb.sury.org+1
Configuration: /var/www/owncloud/apps/password_policy/tests/unit/phpunit.xml
Error:         No code coverage driver is available

..................E..EEE.......................................  63 / 169 ( 37%)
............................................................... 126 / 169 ( 74%)
...........................................                     169 / 169 (100%)

Time: 1.92 seconds, Memory: 16.00MB

There were 4 errors:

1) OCA\PasswordPolicy\Tests\Db\OldPasswordMapperTest::testGetOldPasswords with data set #1 (true)
Doctrine\DBAL\Exception\DriverException: An exception occurred while executing 'SELECT a.* FROM (SELECT * FROM "oc_user_password_history" WHERE ("uid" = ?) AND ("password" <> 'expired') ORDER BY "id" desc) a WHERE ROWNUM <= 3' with params ["test1"]:

ORA-00932: inconsistent datatypes: expected - got CLOB

/var/www/owncloud/lib/composer/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php:76
/var/www/owncloud/lib/composer/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php:184
/var/www/owncloud/lib/composer/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php:158
/var/www/owncloud/lib/composer/doctrine/dbal/lib/Doctrine/DBAL/Connection.php:943
/var/www/owncloud/lib/private/DB/Connection.php:187
/var/www/owncloud/lib/composer/doctrine/dbal/lib/Doctrine/DBAL/Query/QueryBuilder.php:219
/var/www/owncloud/lib/private/DB/QueryBuilder/QueryBuilder.php:141
/var/www/owncloud/apps/password_policy/lib/Db/OldPasswordMapper.php:51
/var/www/owncloud/apps/password_policy/tests/unit/Db/OldPasswordMapperTest.php:97
[...]

PR #114, opened 2018-08-27
Fix OldPasswordMapper for Oracle

PVince81:
@phil-davis can you run the Oracle drone here ?

Not sure on what branch you got the results from https://github.com/owncloud/password_policy/issues/101 ...

Fixes https://github.com/owncloud/password_policy/issues/101

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -196,14 +196,13 @@ matrix:
       NEED_SERVER: true
       NEED_INSTALL_APP: true
 
-# Oracle unit test failures - see issue 103
-#    - PHP_VERSION: 7.1
-#      OC_VERSION: daily-master-qa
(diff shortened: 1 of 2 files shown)


PVince81:
tested locally, to_char doesn't exist for mysql. I'll add an Oracle condition...

PVince81:
added if statement for Oracle

phil-davis:
This was based off some old point of master, so it had no ``.drone.yml`` and no acceptance test code.
I am rebasing to latest master, then will add Oracle drone.

PVince81:
maybe I pushed too fast and forgot "--force" and context-switched too fast.

anyway, I added a new commit with the condition now, let's see

Issue #66, opened 2018-07-17
occ user:expire-password not working, when expiry is not explicitly enabled


patrickjahns:
During smoke testing the release tarball, I noticed a confusion.

After installing and enabling `password_policy` app - the occ command `occ user:expire-password` is available. 
Withou further configuration, I have used the `occ user:expire-password user1` - but confusingly `user1` was still able to login.

Only when checking the top expiry box and using the occ command - the user was forced to change his password on login

![image](https://user-images.githubusercontent.com/6403243/42836087-79eaa276-89fa-11e8-9aaf-8dbb2ecd5e4f.png)

PVince81:
@pmaier1 something to mention in docs or release notes ? known issue ?

PVince81:
also to discuss: what is the expected behavior ?

1) occ command fails "password expiration is not enabled": easy fix
or
2) expiration works anyway: more involved fix as we need to make expiration system work despite expiration being disabled

pmaier1:
1. does not make much sense as the sysadmin could enable it via occ app:config and execute the occ command again, right?

The separation between oC admin and sysadmin is not perfect in this case but to avoid confusion I would go for 2.

PVince81:
Estimate: 0.5-3md depending on what we find

PVince81:
Looking into the code it looks like having password expiration disabled would be the equivalent of having a rule "every password must expire after 0 days" but would need to limit said rule to only virtual password entries created by the admin through the command.

It becomes more complicated if the admin runs the occ command not with the default date (expire now) but expiration in the future.

I'm not too keen on adding this additional layer of complexity in the already complex code paths we have now unless there is a big benefit from it.

Also note that with the approach with "0 days", if the admin decides to tick the checkbox "days until user password expired" the value of 0 days suddenly becomes 90 days (or whatever is in the field), which is equivalent to changing the "days" value, which, according to the docs, requires the admin to run the occ command again for all users.

@pmaier1 did you hope to have this feature mostly for immediate expiry or would future expiry still be interesting despite not having any rule configured ?

PVince81:
A middle-ground approach would be to allow only immediate expiration when no expiration rules were set. This would be easier as we could directly set the "password has expired" flag for said users and would need us to send the notification right away.

PVince81:
Ok, I decide to go the easy route for now: disable the command completely if no expiration rule was configured: https://github.com/owncloud/password_policy/pull/115

pmaier1:
[...]

PR #115, opened 2018-08-27
Disable occ user:expire-password if no expiration rule was configured

PVince81:
Fixes https://github.com/owncloud/password_policy/issues/66

Disables the command completely. See ticket for discussion.


PVince81:
rebased to get drone

Issue #100, opened 2018-08-23
Command should output error in invalid arguments given


PVince81:
See https://github.com/owncloud/password_policy/pull/77#issuecomment-415452503

Run `occ user:password-expire user1` on master where the above PR was merged.

I expected some error to be output instead of nothing at all!

@sharidas FYI

PR #122, opened 2018-09-03
Expire command should let user know about invalid argument

sharidas:
Expire command should let user know if the argument
provided is invalid or not.


sharidas:
Fixes https://github.com/owncloud/password_policy/issues/100

Issue #126, opened 2018-09-04
Automated tests for password_policy (part 2)


phil-davis:
- [x] check password complexity combinations on creating public link shares (done with webUI)
- [x] check password complexity combinations on updating public link shares ~@paurakhsharma is making~ #140
- [x] check number of last passwords - API and UI #134 

Current to do:
- [x] check public link password expiration date limit enable #288 #312 
- [x] check public link password expiration date limit changing the maximum days #312

For later:
- [x] check forcing user password expiration via `occ user:password-expire` command  https://github.com/owncloud/password_policy/pull/314
- [x] check general time-based password expiry #314 #318
- [x] check that required notifications are sent before and at password expiry  #318

The above points need thought about how to "black-box" test the effectiveness of expiration of authentication set for future dates that are well in advance of current time.

- I moved the last 2 points to a separate issue #315

phil-davis:
For now, add test scenarios that set the "Public link expiration policies" and then create public links and confirm that the policy is enforced (you should get some error status when trying to create a public link without an expiration date, or with an expiration date past the maximum).

Also add scenarios that set core "share via link" default expiration date, and enforce that, and make sure that the shorter expiration date is enforced.

phil-davis:
@HariBhandari07 in a new PR please also add scenarios that change the maximum expiry dates.

dpakach:
@phil-davis I think we did everything from current to do list in https://github.com/owncloud/password_policy/issues/126#issue-356814151.
Now remaining part is the for later section.

phil-davis:
- I moved the last 2 points to a separate issue #315 

To finish this issue:
- [ ] add scenario for when the user's password is expired, the user goes to reset their password, and they put in passwords that do not match
- [ ] add scenario for when the user's password is expired, there are some password requirements (e.g. minimum length) the user goes to reset their password, and they put in a new password that does not meet the requirements


(a couple of not-so-happy-path scenarios to know that the reset expired password screen behaves OK)

PR #134, opened 2018-09-19
API and UI tests for last passwords user password policy

phil-davis:
Part of issue #126


phil-davis:
Needs provisioning API password reset acceptance test steps to become available in core QA tarball.

PR #151, opened 2018-10-03
Test refactor 20181003

phil-davis:
Get acceptance tests to pass again with current core.

Details coming...


phil-davis:
This needs the latest testing app code, so it can successfully create a user with display name containing space "User One" by asking the testing app too do an occ command.

PR #166, opened 2018-11-22
Move phpunit.xml to root folder

phil-davis:
- Put ``phpunit.xml`` into the root folder of the app.
- Put the coverage report into ``tests/output/clover.xml``
- Use the ``phpunit`` from core
- Remove excess items from ``require-dev`` of ``composer.json``


phil-davis:
@PVince81 @DeepDiver1975 this seems to be what has been done in other apps.

Is this the current "standard"?

If so, then we can easily adjust other apps the same way.

phil-davis:
@DeepDiver1975 please review and approve.

phpunit.xml
```diff
@@ -1,25 +1,25 @@
 <?xml version="1.0" encoding="utf-8" ?>
-<phpunit bootstrap="bootstrap.php"
+<phpunit bootstrap="tests/unit/bootstrap.php"
```
patrickjahns: @DeepDiver1975 - you used `failOnWarning` recently - should we add it directly here?

phil-davis: I have added it below. Let's see if CI passes.

phil-davis:
@DeepDiver1975 @patrickjahns @PVince81 this is ready for review again.
I got rid of the local ``phpunit`` and it uses ``phpunit`` from core, in the same way that ``data_exporter`` is doing.

phil-davis:
This makes the CI infrastructure in this app more like "the imagined standard".
It implements stuff like what is already in ``data_exporter``

@patrickjahns this one will be easier to review (smaller) than the files_antivirus one. So it might be good to get this agreed and merged first. Then we just check that https://github.com/owncloud/files_antivirus/pull/273 implements things the same way.

.gitignore
```diff
@@ -5,14 +5,15 @@
 # Composer
 composer.phar
 vendor/
+vendor-bin/**/vendor
```
paurakhsharma: when we have `vendor/` in `.gitignore` file, do we really need `vendor-bin/**/vendor`

paurakhsharma: @phil-davis, IMO if we add `/vendor` it will only ignore the top-level directory. But in this case (i.e when using `vendor/`) every `vendor` folder inside the repo is ignored.

phil-davis: We need to understand correctly what happens.
@patrickjahns ?

phil-davis: I tested this just now. What @paurakhsharma says is true. Every folder called ``vendor`` anywhere in the tree is ignored just by writing ``vendor/`` in ``.gitignore``

@patrickjahns please advise what you want to do. Because we need to have a consistent format for this stuff in every repo. Otherwise it is confusing for people who wonder about small differences.

PR #167, opened 2018-11-26
Check that coding standard passes on PHP 5.6

phil-davis:
This will report any syntax issues if there is new 7.* syntax used in the code.
For the moment we need to still use syntax that works on PHP 5.6

I left the PHP 7.2 coding standard check in the matrix. It feels good to know that the code also passes syntax checks on PHP 7.2 - maybe there are things that work in PHP 5.6 and are deprecated and gone by PHP 7.2

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -202,6 +202,9 @@ matrix:
     - PHP_VERSION: 7.2
       TEST_SUITE: owncloud-coding-standard
 
+    - PHP_VERSION: 5.6
+      TEST_SUITE: owncloud-coding-standard
+
     # Unit Tests
     - PHP_VERSION: 7.1
       OC_VERSION: daily-master-qa


phil-davis:
Note: here is a little example of code that passes in PHP 5.6 and fails in PHP7.*
https://github.com/owncloud/password_policy/pull/168

So IMO it is also good to do a code-style check against the latest PHP (in this case 7.2) just to verify that there is not old PHP 5.6 "bad/naughty" syntax that is going to explode in PHP 7.*

.drone.yml
```diff
@@ -202,6 +202,9 @@ matrix:
     - PHP_VERSION: 7.2
```
DeepDiver1975: as commented by @patrickjahns - it is enough to run the code standard once with php5.6

phil-davis: @patrickjahns please confirm that this is what you want.

PR #168, opened 2018-11-26
Invalid octal integer

phil-davis:


lib/Engine.php
```diff
@@ -75,7 +75,7 @@ public function generatePassword() {
 		$length = 20;
 		$password = '';
 		if ($this->yes('spv_min_chars_checked')) {
-			$length = $this->configValues['spv_min_chars_value'] + 1;
+			$length = $this->configValues['spv_min_chars_value'] + 018;
```
phil-davis: Note: In PHP 5.6 this is parsed as an octal ``1`` (= decimal ``1`` = ``1`` in any base). The PHP 5.6 parser sees the ``8`` (not an octal digit) and happily stops there, truncating the numeric value.

In PHP 7.* the parser has a fit when it sees ``8`` or ``9`` when it is trying to parse an octal number.

In this case ``owncloud-coding-standard`` and ``phpunit`` tests pass on PHP 5.6 and fail on PHP 7.2. So this is why I think it is useful to have ``owncloud-coding-standard`` and ``phpunit`` tests running on PHP 5.6 and on whatever is the most current PHP supported.

PR #169, opened 2018-11-26
Implement phpstan and phan

phil-davis:
- implement phan in drone CI
- make code changes so that phan is happy
- drop coding-standard in PHP7.2, phan will now be an effective lint check for PHP7.* and coding-standard is running in PHP5.6 which also provides a lint check there
- implement phpstan infrastructure so it can be run locally
- add phpstan step to drone, but do not make any matrix entry for it, so that it does not actually run in CI

A followup PR can sort out how to hack the code so that ``phpstan`` passes.


phil-davis:
``phan`` does find syntax errors. When I make a change like in https://github.com/owncloud/password_policy/pull/168/files#diff-2ce8f556409a432e7318b01ff99e11c8R78 then:
```
lib/Engine.php:78 PhanSyntaxError Invalid numeric literal
```
So that is good - it is being a ``lint`` for us on PHP7.*

phil-davis:
@patrickjahns @DeepDiver1975 this is at point where I think it is useful to review/merge.
Then "we" (=someone) can sort out making ``phpstan`` pass in another PR.

.drone.yml
```diff
@@ -205,6 +229,28 @@ matrix:
     - PHP_VERSION: 5.6
       TEST_SUITE: owncloud-coding-standard
 
+    # phan
+    - PHP_VERSION: 7.2
```
phil-davis: Question from me, the uninitiated.
Run ``phan`` on all PHP 7.* versions? Or just one of them? And which one?

patrickjahns: phan needs to run on the version we want to check against, as it requires php to convert it to its internals to check the AST.

phan thus should be run with every php version we support ( except 5.6 which does not have AST )

lib/UserNotificationConfigHandler.php
```diff
@@ -98,7 +98,12 @@ public function getExpirationTimeForNormalNotification() {
 	 * to include the userid owning the password and an id for the password
 	 */
 	public function markAboutToExpireNotificationSentFor(OldPassword $passInfo) {
-		$this->config->setUserValue($passInfo->getUid(), 'password_policy', 'aboutToExpireSent', $passInfo->getId());
+		$this->config->setUserValue(
```
patrickjahns: Please share the error which is fixed through this

lib/UserNotificationConfigHandler.php
```diff
@@ -108,7 +113,12 @@ public function markAboutToExpireNotificationSentFor(OldPassword $passInfo) {
 	 * to include the userid owning the password and an id for the password
 	 */
 	public function markExpiredNotificationSentFor(OldPassword $passInfo) {
-		$this->config->setUserValue($passInfo->getUid(), 'password_policy', 'expiredSent', $passInfo->getId());
+		$this->config->setUserValue(
```
patrickjahns: Please share the error which is fixed through this

lib/Jobs/PasswordExpirationNotifierJob.php
```diff
@@ -192,7 +192,7 @@ private function sendPassExpiredNotification(OldPassword $passInfo, $expirationT
 		$notification->setApp('password_policy')
 			->setUser($passInfo->getUid())
 			->setDateTime(new \DateTime("@{$notificationTimestamp}"))
-			->setObject('expired', $passInfo->getId())
+			->setObject('expired', (string)$passInfo->getId())
```
patrickjahns: Please share the error which is fixed through this

lib/Jobs/PasswordExpirationNotifierJob.php
```diff
@@ -147,7 +147,7 @@ private function sendAboutToExpireNotification(OldPassword $passInfo, $expiratio
 		$notification->setApp('password_policy')
 			->setUser($passInfo->getUid())
 			->setDateTime(new \DateTime("@{$notificationTimestamp}"))
-			->setObject('about_to_expire', $passInfo->getId())
+			->setObject('about_to_expire', (string)$passInfo->getId())
```
patrickjahns: Please share the error which is fixed through this

lib/Db/OldPasswordMapper.php
```diff
@@ -39,7 +39,7 @@ public function __construct(IDBConnection $db) {
 	 * @return OldPassword[]
 	 */
 	public function getOldPasswords($uid, $length, $excludeForceExpired = true) {
-		/* @var IQueryBuilder $qb */
+		/* @var OCP\DB\QueryBuilder\IQueryBuilder $qb */
```
patrickjahns: Please share the error which is fixed through this

lib/Db/OldPassword.php
```diff
@@ -26,12 +26,12 @@
 
 /**
  * @method int getId()
- * @method void setId(\int $id)
+ * @method void setId(int $id)
```
patrickjahns: Please share the error which is fixed through this

lib/UserNotificationConfigHandler.php
phil-davis: ``$passInfo->getId()`` returns an ``int`` but ``setUserValue()`` expects a string.

lib/UserNotificationConfigHandler.php
phil-davis: ``$passInfo->getId()`` returns an ``int`` but ``setUserValue()`` expects a string.

lib/Jobs/PasswordExpirationNotifierJob.php
phil-davis: ``$passInfo->getId()`` returns an ``int`` but ``setObject()`` expects a string.

lib/Jobs/PasswordExpirationNotifierJob.php
phil-davis: ``$passInfo->getId()`` returns an ``int`` but ``setObject()`` expects a string.

lib/Db/OldPassword.php
phil-davis:
```
lib/Db/OldPassword.php:28 PhanUndeclaredTypeParameter Parameter $id has undeclared type \int
lib/Db/OldPassword.php:29 PhanParamSignatureMismatch Declaration of function setId(\int $id) : void should be compatible with function setId(int $id) : void defined in ../../lib/public/AppFramework/Db/Entity.php:28
```

.drone.yml
phil-davis: pushed a change for this

lib/Db/OldPasswordMapper.php
patrickjahns: > Somehow phan seemed to stupid to understand that the use statement was "used" by the PHPdoc for the $qb variable.

Not a problem within phan - phan analysies the AST ( after sourcecode has been convered ). For performance reasons, the AST does not have inline comments for every machine code - it would make it too slow. 
Comments are available on the class // method level

phil-davis: So in this case the change here is the "correct" solution?
Because PHP execution has no need for the ``use`` statement. It was only conventions of PHPdoc parsers that were parsing the ``use`` statement for convenience to work out what ``IQueryBuilder`` meant.

But in (unusual?) cases where there are many ``@var OCP\DB\QueryBuilder\IQueryBuilder`` PHPdoc throughout a class, but no actual real use of it, then there will be duplication of ``OCP\DB\QueryBuilder\`` in multiple PHPdoc blocks.

.drone.yml
```diff
@@ -254,7 +254,8 @@ matrix:
       TEST_SUITE: owncloud-coding-standard
 
     # phan
-    - OC_VERSION: daily-master-qa
+    - PHP_VERSION: 7.1
```
phil-davis: @patrickjahns I need to have "server" code available so that ``phan`` knows about the core code/classes etc.
When I put ``NEED_SERVER`` for this matrix entry, then the ``NEED_SERVER`` pipeline step needs ``PHP_VERSION`` to be set.

When the first ``phan-70`` pipeline step runs, it does ``composer install`` and gets the needed ``phan`` dependencies... and that is all good.

When ``phan-71`` and ``phan-72`` run, they do not go through a "full" ``composer install`` 

It all passes, but I am not that confident about whether we are really doing effective tests for each of PHP 7.0 7.1 and 7.2

An alternative is to go back to having 3 matrix entries, 1 for ``phan`` on each PHP version. That will ensure the server and ``phan`` and everything are really using the requested PHP version.

Thoughts?

patrickjahns: The dependencies from phan and core are completely segregated. That is what `composer-bin` ensures.
As phan itself is having a platform limited to php 7.0 ( https://github.com/phan/phan/blob/master/composer.json#L20-L22 ) - we are fine either way on installing it on any php version > 7.0

Phan will evaluated the AST from the code in question in the php runtime it currently is executed. So we need to execute it with every php version we want it to be evaluated against

patrickjahns: > Phan will evaluated the AST from the code in question in the php runtime it currently is executed. So we need to execute it with every php version we want it to be evaluated against

Adding to that - the core dependencies should also not change per php version - so it does not matter. And as stable10 is also php7.2 compatible - we might always use php7.2 for the core install step ( no matter if it is master or stable10 )

phil-davis: Bumped PHP_VERSION here from 7.1 to 7.2

phil-davis:
If CI passes, then I think this is ready @patrickjahns

.drone.yml
patrickjahns: I was wrong - `7.1` is/was fine ( thought that master required min 7.2 🙈 ). 
But `7.2` doesn't hurt either way - it shouldn't make a difference in this matter

phil-davis:
CI passes.
@patrickjahns I think all the code comments have been addressed.
Please review again.

phil-davis:
@patrickjahns ping for review

phil-davis:
@patrickjahns I rebased and started getting a fail of ``phan`` on PHP 7.0. PHP 7.0 is not supported for core ``master`` so it always has the chance to fail of something in core ``master`` is not agreeable with some PHP 7.0 related thing.

I push another commit to run ``phan`` for PHP 7.0 using core ``stable10`` and that passes.

Please review and give your comments or approval.

phil-davis:
@patrickjahns ping for review

Issue #159, opened 2018-11-08
Password Policies should not affect to LDAP users login password


davitol:
### Steps to reproduce

1 - As admin configure an LDAP backend
2 - As admin set User password policies: Force users to change their password on first login
3 - Try to log in with an LDAP User
4 - Set a new password window appears
 

### Expected behaviour

Password expiration should not affect to LDAP users

### Actual behavior

LDAP user cannot login anymore

### Server configuration

**Operating system**:
Ubuntu 16.04

**Web server:**
Apache

**Database:**
MySQL

**PHP version:**
7.1

**ownCloud version:**

ownCloud 10.0.10

**Updated from an older ownCloud or fresh install:**
Fresh

**Are you using external storage, if yes which one:** 
No

**Are you using encryption:** 
No
**Logs**

```

```

### Client configuration

**browser**

Firefox

<img width="534" alt="screen shot 2018-11-08 at 08 41 15" src="https://user-images.githubusercontent.com/10596560/48185129-678e8c80-e334-11e8-867c-40d9a3445df7.png">

![screen shot 2018-11-08 at 08 41 53](https://user-images.githubusercontent.com/10596560/48185133-6c534080-e334-11e8-8f77-5f5afe749211.png)

(cc @pmaier1)

PVince81:
I remember we had checks in place to see if the user backend supports setting passwords.

In this case we need to make sure such check is in place for all code paths that enforce rules.

pmaier1:
Yes, please schedule this one. Ugly side-effects.. :)

PVince81:
@sharidas can you have a look and estimate ? I think we had something similar in another code path

sharidas:
Looking at this problem I feel we can have a slightly modified version of https://github.com/owncloud/password_policy/pull/171. That is on top of this change we can enforce password policy to apply only for user backendclassname `Database`. I did a minor modification to the pr as shown below:
```
if ($user->getBackendClassName() === 'Database') {
		// try to find user in account table, needs find to search additional search terms,
		$this->forcePasswordChange(true, $user);
}
```

With the above change set and after enabling `Force users to change their password on first login` in password_policy, I was able to acheive the points mentioned below:
- [x] ldap users were able to login and password policy's password reset form was not visible.
- [x] for users who were created using password from the user_management, were able to see password reset form from password_policy.

I hope I am not missing anything here. Any comments?

sharidas:
[...]

PR #173, opened 2018-11-28
Ignore ldap user for password change during first login

sharidas:
LDAP users can be ignored for the password change
during first time login.


sharidas:
Fixes https://github.com/owncloud/password_policy/issues/159

This change helps ignore the LDAP users when password change is enforced for first time login. It checks the users backendclass name to check if it is `Database`. For LDAP users the backend class name is `LDAP`.

PR #180, opened 2018-11-30
Change paths to files in drone codecov step

phil-davis:
It looks odd to me writing a full file name in ``paths:`` (although it does work)
It feels nicer to put a specific codecov file in ``files:``

Change it here before we copy this in too many other places.


phil-davis:
we went a different way

PR #183, opened 2018-12-06
Remove unneeded phpunit Makefile dependencies

phil-davis:
We run the ``phpunit`` that already comes "for free" in core.
The acceptance tests do not use ``phpunit`` and do not depend on any special dev/test framework code being installed in the app.
So neither of these need to depend on ``vendor/bin/phpunit`` - in fact nothing does.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,12 +104,12 @@ clean: clean-deps clean-dist clean-build
 
 .PHONY: test-php-unit
 test-php-unit:             ## Run php unit tests
-test-php-unit: vendor/bin/phpunit
+test-php-unit:
 	$(PHPUNIT) --configuration ./phpunit.xml --testsuite unit
 
 .PHONY: test-php-unit-dbg
 test-php-unit-dbg:         ## Run php unit tests using phpdbg
-test-php-unit-dbg: vendor/bin/phpunit
+test-php-unit-dbg:
 	$(PHPUNITDBG) --configuration ./phpunit.xml --testsuite unit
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -134,17 +134,17 @@ test-php-phpstan: vendor-bin/phpstan/vendor
...
 .PHONY: test-acceptance-cli
 test-acceptance-cli:       ## Run CLI acceptance tests
-test-acceptance-cli: vendor/bin/phpunit
+test-acceptance-cli:
 	../../tests/acceptance/run.sh --remote --type cli
 
 .PHONY: test-acceptance-webui
 test-acceptance-webui:     ## Run webUI acceptance tests
-test-acceptance-webui: vendor/bin/phpunit
+test-acceptance-webui:
```
skshetry: After removing the dependency, new/updated packages are not installed, right? So, tests might fail locally if they depend on external packages?

Currently, though the app does not have any `dev` dependencies.

phil-davis: It won't do whatever it did to try and "make" ``vendor/bin/phpunit`` - and actually we do not care about that. But yes, if the app had some ordinary "require" dependencies in ``composer.json`` then I suppose every ``make test*`` target should end up doing ``composer install`` to make those dependencies get loaded. So we do need to think about it in other apps, and put some sensible dependency.

In this case, ``composer install`` in the app will get the ``vendor-bin`` stuff, which is coding-standard, phan, phpstan. And we do not need any of those to run acceptance tests.

PR #184, opened 2018-12-08
[do not merge] Trigger CI

phil-davis:
to see if there is any problem


phil-davis:
Last night there was:
https://drone.owncloud.com/owncloud/password_policy/376/474
```
SCENARIO RESULT: (fail)
  Scenario Outline: administrator creates a user with password set to a string that has too few uppercase letters                                                                         # /var/www/owncloud/apps/password_policy/tests/acceptance/features/webUIPasswordAddUser/addUserUppercaseLetters.feature:24
    When the administrator attempts to create a user with the name "guiusr1" and the password "<password>" using the webUI                                                                # WebUIUsersContext::theAdminCreatesAUserUsingTheWebUI()
    Then a notification should be displayed on the webUI with the text "Error creating user: The password contains too few uppercase letters. At least 3 uppercase letters are required." # WebUIGeneralContext::aNotificationShouldBeDisplayedOnTheWebUIWithTheText()
    And user "guiusr1" should not exist                                                                                                                                                   # FeatureContext::userShouldNotExist()

    Examples:
      | password       |
      | 0uppercase     |
        Failed asserting that two strings are equal.
        --- Expected
        +++ Actual
        @@ @@
        -'Error creating user: The password contains too few uppercase letters. At least 3 uppercase letters are required.'
        +'There were problems with the code integrity check. More information… Error creating user: The password contains too few uppercase letters. At least 3 uppercase letters are required.'
SCENARIO RESULT: (fail)
      | Only2Uppercase |
        Failed asserting that two strings are equal.
        --- Expected
        +++ Actual
        @@ @@
        -'Error creating user: The password contains too few uppercase letters. At least 3 uppercase letters are required.'
        +'There were problems with the code integrity check. More information… Error creating user: The password contains too few uppercase letters. At least 3 uppercase letters are required.'
```
Other test scenarios passed before any after that. So how could there be a transient "code integrity check" appear for just a minute?

PVince81:
The integrity check message appears in the notification box. Since the UI tests seem to be parsing it in whole, you'll get both error messages.

Now the mystery is where the integrity check error comes from... probably not the password policy app as we're not installing from tarball. So likely the stable10 qa one ?

phil-davis:
I tried making some long-running webUI tests in the notifications app, and was hoping that those would start failing after 30 minutes. But they did not. But notifications app has the notifications displayed in a different way. I think the tests will fail if the number of notifications is wrong.

I will see if I can easily make a PR in core that has webUI tests for ordinary notifications, using the same step code that is used in password policy. And make it long-running, and see what happens.

I raised this because it would be nice to understand why the integrity check notification has started coming - did something change in the way the tarball builds, or something else in core or???

PR #185, opened 2018-12-08
[do not merge] Test web ui password add user

phil-davis:


phil-davis:
This fails webUI acceptance tests for all PHP 5.6 7.0 and 7.1 because after 30 minutes the integrity check message starts happening.

PR #187, opened 2018-12-10
More flexible Makefile help column format

phil-davis:
When formatting the ``make help`` output:
- replace multiple spaces by a single space
- use the ``column`` command to line up the ``Makefile`` targets and descriptions into columns

At the moment we have to get the correct number of spaces in each ``test-xyz:        ## Description of command`` line.

After this change, we can put whatever spaces we like before and after the ``##`` and the ``make help`` still lines up.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ PHPSTAN=php -d zend.enable_gc=0 vendor-bin/phpstan/vendor/bin/phpstan
 
 # start with displaying help
 help:
-	@fgrep -h "##" $(MAKEFILE_LIST) | fgrep -v fgrep | sed -e 's/\\$$//' | sed -e 's/##//'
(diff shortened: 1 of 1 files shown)


phil-davis:
Inspired by @skshetry doing similar in ransomware_protection.

skshetry:
Would it be too much if I ask for colored target name? 
ref: https://gist.github.com/prwhite/8168133#gistcomment-976470

Makefile
```diff
@@ -100,50 +100,50 @@ clean: clean-deps clean-dist clean-build
 
 ##
```
skshetry:
```suggestion
##---------------------
```
We need this to separate into multiple sections. Otherwise, it will be very hard to read.

phil-davis:
The colour feature request does not immediately work in my terminal. It will need effort to make sure it will work nicely for all (reasonable) terminals that developers use with *nix and Mac. That can be a weekend PR for "someone who cares".

phil-davis:
Rebased to trigger CI again.

PR #191, opened 2018-12-13
Check drone CI

phil-davis:


phil-davis:
drone was broken for a time.

PR #194, opened 2018-12-20
tests of guest user setting password

individual-it:
test if the password policy is obeyed when guest user sets own password
depends on https://github.com/owncloud/guests/pull/298


individual-it:
@phil-davis fixed the issues

PR #201, opened 2019-01-19
Trigger CI

phil-davis:
To see if Oracle phpunit fails like in https://drone.owncloud.com/owncloud/update-testing/298/86


phil-davis:
Oops - that phpunit test fail is in update-testing nd related to configreport - nothing to do with password policy!

PR #202, opened 2019-01-22
Add php_codesniffer to check acceptance test code style

phil-davis:
Like was merged in core in https://github.com/owncloud/core/pull/34191

So that we know if we break our own rules for acceptance test code style.


individual-it:
any reason not to use the phpcs.xml file from core? so we don't need to edit it in multiple places

phil-davis:
> any reason not to use the phpcs.xml file from core? so we don't need to edit it in multiple places

the QA tarball does not have phpcs.xml in it

If we moved phpcs.xml in core into the ``tests`` folder then it will "magically" appear in the QA tarball. Since we use it only for acceptance test code, maybe it is [OK|good] to hide it in there anyway?

individual-it:
> If we moved phpcs.xml in core into the tests folder then it will "magically" appear in the QA tarball. Since we use it only for acceptance test code, maybe it is [OK|good] to hide it in there anyway?

no, because it might be used for other files also
it should go wherever the php-cs-fixer config lives

PR #204, opened 2019-02-05
Remove wrongly-named WebUIUsersContext

phil-davis:
When running webUI acceptance tests locally, ``behat`` told me that it did not know about ``WebUIUsersContext``. That is true, because that context in core is called ``WebUIUserContext``

I have no idea why ``behat`` does not complain about this in drone. Anyway the steps in ``WebUIUserContext`` are not needed anyway.

While touching this, sort the contexts so it is easier to look through the list.

### DIFF
--- a/tests/acceptance/config/behat.yml
+++ b/tests/acceptance/config/behat.yml
@@ -8,37 +8,37 @@ default:
         - '%paths.base%/../features/apiPasswordAddUser'
       contexts:
         - PasswordPolicyContext:
-        - OccContext:
         - FeatureContext: &common_feature_context_params
             baseUrl:  http://localhost:8080
             adminUsername: admin
             adminPassword: admin
             regularUserPassword: 123456
             ocPath: apps/testing/api/v1/occ
+        - OccContext:
 
     apiPasswordChange:
       paths:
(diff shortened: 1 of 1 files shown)


tests/acceptance/config/behat.yml
```diff
@@ -8,128 +8,127 @@ default:
...
-        - WebUIGeneralContext:
-        - WebUILoginContext:
         - EmailContext:
+        - FeatureContext: *common_feature_context_params
         - WebUIFilesContext:
+        - WebUIGeneralContext:
+        - WebUILoginContext:
+        - WebUIPasswordPolicyContext:
         - WebUIPersonalGeneralSettingsContext:
-        - WebUIUsersContext:
```
phil-davis: This exists in core ``stable10`` and ``webUIPasswordAddUser`` suite is just run against core ``stable10`` so it can test user_management stuff with password_policy.
Some day if user_management is separated out as a separate app in ``stable10`` then this will need to be reviewed.
But for now this is the way it should work.

phil-davis:
password_policy ``webUIPasswordAddUser`` needs ``WebUIUsersContext`` for testing user management functionality against core ``stable10``
We can't remove ``WebUIUsersContext``

PR #207, opened 2019-03-01
[poc] Use common makefiles

phil-davis:
Notes:
- to run just owncloud-coding-standard this has to set `NEED_CORE: true` so it can find the make file(s) to include, so it will know how to `make test-php-style` (other analysis like `phan` or `phpstan` needs core anyway, so it can find and parse the common core code/classes that an app includes)


phil-davis:
Closing see comment https://github.com/owncloud/richdocuments/pull/253#issuecomment-513429624

PR #213, opened 2019-03-18
disable integrity check for acceptance tests

### LINKED ISSUE #186: acceptance tests fail after 30 minutes with integrity check messages
See PR #184 
After 30 minutes and integrity check seems to "go off". After that, there is an extra notification in the webUI for the administrator - ``There were problems with the code integrity check`` - This started happening last Friday night.

There happen to be acceptance tests that are checking for notifications about the password not meeting the policy. These fail because of the extra notification that starts happening 30 minutes into the test run.

Note: the fail in PR #184 is only in PHP 5.6 - because that runs slow. PHP 7.* runs faster and finishes in less than 30 minutes.

PR #185 makes the test scenarios go longer than 30 minutes. It demonstrates that the problem happens just the same for all PHP versions.

We can easily adjust the tests to ignore other "bonuses" in the notifications.

But why do we seem to now be getting "integrity check" errors?
Is it an integrity check problem in the password policy code, or the core stable10 QA tarball?

@PVince81 @individual-it

individual-it:
disable integrity because of owncloud/core#34787
fixes #186

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -29,6 +29,7 @@ pipeline:
       - php occ a:l
       - php occ config:system:set trusted_domains 1 --value=owncloud
       - php occ log:manage --level 0
+      - php occ config:system:set --value true --type boolean integrity.check.disabled
     when:
       matrix:
         NEED_INSTALL_APP: true


individual-it:
seems to work see https://drone.owncloud.com/owncloud/password_policy/558/504

individual-it:
an other reason to fail the check is the app itself  see https://github.com/owncloud/QA/issues/614

PR #217, opened 2019-04-21
Cast times to int (phan warning)

phil-davis:
phan 1.3.0 started warning:
https://github.com/phan/phan/issues/2656
https://drone.owncloud.com/owncloud/password_policy/605/82
```
lib/Notifier.php:76 PhanTypeInvalidLeftOperandOfAdd Invalid operator: left operand of + is string (expected array or number)
lib/Notifier.php:76 PhanTypeInvalidRightOperandOfAdd Invalid operator: right operand of + is string (expected array or number)
Makefile:134: recipe for target 'test-php-phan' failed
make: *** [test-php-phan] Error 1
```

According to our PHPdoc `$notification->getMessageParameters()` returns an array of strings. We are taking elements from the array and adding them together as integers to calculate the expiration time (in integer seconds). So cast to `int` before adding.

### DIFF
--- a/lib/Notifier.php
+++ b/lib/Notifier.php
@@ -71,8 +71,8 @@ class Notifier implements INotifier {
 		$messageParams = $notification->getMessageParameters();
 		$currentTime = $this->timeFactory->getTime();
 		$currentDateTime = new \DateTime("@{$currentTime}");
-		$passwordTime = $messageParams[0];
-		$expirationTime = $messageParams[1];
+		$passwordTime = (int) $messageParams[0];
+		$expirationTime = (int) $messageParams[1];
 		$targetExpirationTime = $passwordTime + $expirationTime;
 		$expirationDateTime = new \DateTime("@{$targetExpirationTime}");
 		$interval = $currentDateTime->diff($expirationDateTime);
--- a/lib/Notifier.php
+++ b/lib/Notifier.php
@@ -71,6 +71,10 @@ class Notifier implements INotifier {
 		$messageParams = $notification->getMessageParameters();
 		$currentTime = $this->timeFactory->getTime();
 		$currentDateTime = new \DateTime("@{$currentTime}");
+		// getMessageParameters() returns strings. In this case the strings are
+		// decimal integers representing the time (in seconds) when the password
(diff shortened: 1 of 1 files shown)


lib/Notifier.php
```diff
@@ -71,8 +71,8 @@ private function formatAboutToExpire(INotification $notification, IL10N $l) {
 		$messageParams = $notification->getMessageParameters();
 		$currentTime = $this->timeFactory->getTime();
 		$currentDateTime = new \DateTime("@{$currentTime}");
-		$passwordTime = $messageParams[0];
-		$expirationTime = $messageParams[1];
+		$passwordTime = (int) $messageParams[0];
+		$expirationTime = (int) $messageParams[1];
```
phil-davis: What if something passes in an `INotification` that has message parameters that are wrong, not integer.
There never was any code to check for that. And anyway what to do if the expected time parameters passed in inside the `INotification` are not sensible integers?

jvillafanez: The `getMessageParameters` is expected to return an array of strings. If an array of integers are being sent, that's something to be fixed in whatever code is sending those parameters. Setting parameters other than strings is wrong.

For the string -> integer conversion, I guess we can assume that both parameters can be converted to an integer. If so, we can assume that both time strings will fit into a integer. I guess casting is fine, not sure if `intval()` can provide any advantage here. We can consider this as part of the comunication protocol between the sender and the receiver, so if the sender sends something that can't be converted to an integer, I expect a crash, and the code should crash. This is something that can be ensured by the sender, it can be fixed at development time, and it shouldn't happen in production, that's why I prefer to let the code crash in this case.
My only worry is that it could happen an integer overflow when we add both values (just below). Maybe we can just leave this as an edge case.

In any case, I think it's a good idea to leave some comments with the expected behaviour of the conversion, specially for weird cases.

phil-davis: These days all "real" PHP implementations are 64-bit. The times here are "current" Unix timestamps, and the password expiry seconds (which will be some "small" number). The sun wil have burnt out long before we overflow here ;)
I added a short essay about the (int) casting.

PR #223, opened 2019-05-29
Split PasswordAddUser and PasswordChange suites

phil-davis:
These suites were taking quite a long time:
apiPasswordAddUser 53m
apiPasswordChange 60m
webUIPasswordAddUser 60m
webUIPasswordChange 45m

Split them so that the elapsed time is going going over or near 1 hour.


individual-it:
TODO:
- [ ]  make same changes in user_ldap

PR #234, opened 2019-07-05
Special characters provided shouldn't be empty

sharidas:
The checkbox should not allow special characters
to be empty.

### DIFF
--- a/js/ajax.js
+++ b/js/ajax.js
@@ -22,6 +22,13 @@ $(document).ready(function() {
 		event.preventDefault();
 		OC.msg.startSaving('#password_policy .msg');
 
+		if (($("input[name=spv_def_special_chars_checked]").is(':checked') === true) &&
+			($("input[name=spv_def_special_chars_value]").val() === "")) {
+			OC.Notification.showTemporary(t('password_policy', 'Error: The special characters cannot be empty.'));
+			OC.msg.finishedSaving('#password_policy .msg', {status: "failed", data: {message: t('password_policy', 'Failed to save!')} });
+			return;
+		}
+
 		// Load the new token
 		$.ajax({
 			type: "POST",
--- a/tests/acceptance/features/webUIPasswordPolicySettings/setPolicySettings.feature
+++ b/tests/acceptance/features/webUIPasswordPolicySettings/setPolicySettings.feature
@@ -68,6 +68,15 @@ Feature: set password policy settings
     And restrict to these special characters should be set to "!@#$%^&*"
     And the restrict to these special characters password policy should be enabled
 
+  Scenario: set "restrict to these special characters" empty receives user notification and the change is not saved
+    When the administrator enables the restrict to these special characters password policy using the webUI
+    And the administrator sets the restricted list of special characters to "" using the webUI
+    And the administrator saves the password policy settings using the webUI
+    Then notifications should be displayed on the webUI with the text
+      | Error: The special characters cannot be empty. |
+    And the administrator reloads the admin security settings page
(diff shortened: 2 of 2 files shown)


phil-davis:
If I go to Settings->Admin->Security and check "1 special characters" but do not check "Restrict to these special characters" then when I try to save it says:
```
Error: The special characters cannot be empty.
```
But actually this should work.

When "Restrict to these special characters" is not checked, then the text box can be empty. And actually the system implements a default set of special characters - which is all non-alphanumeric characters.

Also, if I do not even check "1 special characters" then I still get the error message.

sharidas:
## Description
When the user checks `Restrict to these special characters:` and provides empty data, the user creation fails. And hence this change set takes care of the same. I have made changes to the controller as well as the frontend. In the front end care has been taken so that the failure notification would be shown in the UI when the user provides emtpy data to `Restrict to these special characters:` when the checkbox before the label is checked. If it is not checked then notification will not be seen.

## Issue
- Fixes https://github.com/owncloud/core/issues/35743

## Test done
- Check the `Restrict to these special characters:` and provide emtpy data and click the save button here is the screenshot what user would see: 
pp
- Refresh the page and the previous value in the text box will be available.
- If the checkbox for `Restrict to these special characters:` is not checked no failure or notification is seen in the page: 
pp2
- The change notification or failure is dependant on 2 factors :
   - If the checkbox `Restrict to these special characters` is checked and
   - If the input text for `Restrict to these special characters` is empty

lib/Controller/SettingsController.php
```diff
@@ -92,6 +93,9 @@ public function updatePolicy() {
 						$value = $this->$convertFuncName($value);
 					}
 					$this->config->setAppValue('password_policy', $key, $value);
+				} elseif (($this->request->getParam('spv_def_special_chars_checked') === 'on') &&
+					($key === 'spv_def_special_chars_value') && ($this->request->getParam($key) === '')) {
+					throw new PolicyException("The restricetd special characters cannot be empty");
```
phil-davis:
```suggestion
					throw new PolicyException("The restricted special characters cannot be empty");
```

sharidas: Done. Corrected to `restricted`.

js/ajax.js
```diff
@@ -22,6 +22,11 @@ $(document).ready(function() {
 		event.preventDefault();
 		OC.msg.startSaving('#password_policy .msg');
 
+		if (($("input[name=spv_def_special_chars_checked]").is(':checked') === true) &&
+			($("input[name=spv_def_special_chars_value]").val() === "")) {
+			OC.Notification.showTemporary(t('password_policy', 'Error: The special characters cannot be empty.'));
```
karakayasemi:
````suggestion
			OC.Notification.showTemporary(t('password_policy', 'Error: The special characters cannot be empty.'));
 			OC.msg.finishedSaving('#password_policy .msg', {status: "failed", data: { message: t('password_policy', 'Failed to save!')} });
 			return;
````
If we can detect an error in client side, I guess no need to post erroneous request to server.  
Also, currently, we are saving `spv_def_special_chars_checked` config value even we say `Failed to save`. Suggested change prevents this situation also .

sharidas: I was trying to be in the defensive side. In either way the data shouldn't be saved, be it from the client side post or any hackish thing which reaches the server.

sharidas: Made changes as discussed.

phil-davis:
Good stuff. I tried to break this just now, and I can't break it.

PR #237, opened 2019-07-19
Revert "Adjust cli tests to no longer expect text in stderr"

sharidas:
Reverts owncloud/password_policy#227


sharidas:
The reason for revert is because of the discussion here https://github.com/owncloud/core/pull/35543#issuecomment-512683143 and the changes merged in core: https://github.com/owncloud/core/pull/35873

PR #242, opened 2019-08-13
Adjust skipOnOcV10.2 test scenarios

phil-davis:
1) Remove the `skipOnOcV10.2` tags
2) Fixup scenario description in occAddUserMinimumLength.feature
3) Make separate scenarios for 10.2 and 10.3 occ command error output to demonstrate the different behaviour of the `occ` command error output in core `10.2` and `10.3`

This "documents" the different behaviour against core `10.2` and the current core `master` (upcoming `10.3`)

Issue https://github.com/owncloud/core/issues/36019 discusses the change in command error message output (among other things). After that is discussed and decided, then there might be further changes depending on what is finally decided for 10.3.

### DIFF
--- a/tests/acceptance/features/cliPasswordChange/occResetPasswordLastPasswords.feature
+++ b/tests/acceptance/features/cliPasswordChange/occResetPasswordLastPasswords.feature
@@ -12,14 +12,12 @@ Feature: enforce the number of last passwords that must not be used when resetti
       | username | password |
       | user1    | Number1  |
 
-  @skipOnOcV10.2
   Scenario: admin resets the password of a user to the existing password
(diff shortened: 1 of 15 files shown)


tests/acceptance/features/cliPasswordChange/occResetPasswordLastPasswords.feature
```diff
@@ -12,14 +12,12 @@ Feature: enforce the number of last passwords that must not be used when resetti
       | username | password |
       | user1    | Number1  |
 
-  @skipOnOcV10.2
```
phil-davis: Actually we do not need to skip these on core `10.2` - the output comes the same to `stderr` on both `10.2` and `10.3`

PR #251, opened 2019-08-28
always exclude the current app when doing install-core

phil-davis:


phil-davis:
PoC only, to make sure that we can exclude in the case when excluding an app/files that are not there anyway.

PR #265, opened 2019-11-15
Dutch translation for 2.0.2

freakynl:
Added nl.js and nl.json for Dutch translation.


DeepDiver1975:
Thanks a lot! Amazing! Unfortunately we cannot accept this that way. All translations are managed on https://www.transifex.com/owncloud-org/owncloud/translate/#nl/password_policy/176538867

Can I ask you to join transifex and update the translations there?

Thanks a lot!

PR #284, opened 2020-03-03
Rebased 2.1.1

phil-davis:
This is a rebased verion of the `release-2.1.1` branch.

Actually I don't think we will have to do it this way. I have cherry-picked stuff into the `release-2.1.1` branch now, and it looks like it will be happy.


phil-davis:
not needed

PR #297, opened 2020-04-16
[Tests-Only] Update phpstan to latest version

dpakach:
Related - owncloud/QA#642

Update phpstan to 0.12

### DIFF
--- a/lib/Command/ExpirePassword.php
+++ b/lib/Command/ExpirePassword.php
@@ -179,7 +179,7 @@ class ExpirePassword extends Command {
 
 			if (\count($uids) > 0) {
 				foreach ($uids as $uid) {
-					/** @var $user \OCP\IUser */
+					/** @var \OCP\IUser|null $user */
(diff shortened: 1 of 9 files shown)


lib/Command/ExpirePassword.php
```diff
@@ -124,7 +124,6 @@ protected function configure() {
 	 *
 	 * @param InputInterface $input
 	 * @param OutputInterface $output
-	 * @return int
```
phil-davis, line 127: execute() should return an `int` status. It already has `return 1` `return 2`  for some error conditions.
IMO there should be a `return 0` at the end of the function.

lib/Engine.php
```diff
@@ -106,7 +106,7 @@ public function generatePassword() {
 	}
 
 	/**
-	 * @param string $password
+	 * @param string|null $password
 	 * @param string $uid
```
phil-davis, line 110:
```suggestion
	 * @param string|null $uid
```
this can also be null?

Issue #315, opened 2020-08-28
Automated tests for password_policy (part 3)


phil-davis:
- [ ] check general time-based password expiry *
- [ ] check that required notifications are sent before and at password expiry *

The above points need thought about how to "black-box" test the effectiveness of expiration of authentication set for future dates that are well in advance of current time.

We generally need to think about how to do this for various features of both apps and core that expire things in the future - e.g. share expiration. Because it would be nice to be more confident that expiring things do actually expire.

PR #318, opened 2020-09-02
[Tests-Only] Added update password UI tests to check password requirements

kiranparajuli589:
## Description
With this PR
- Added password update UI tests cases to check different password requirements
## Related Issue
- Fixes https://github.com/owncloud/password_policy/issues/126
- Fixes https://github.com/owncloud/password_policy/issues/315

### DIFF
--- a/tests/acceptance/features/lib/UpdatePasswordPage.php
+++ b/tests/acceptance/features/lib/UpdatePasswordPage.php
@@ -1,0 +1,142 @@
+<?php
+
+/**
+ * ownCloud
+ *
+ * @author Kiran Parajuli <kiran@jankaritech.com>
+ * @copyright Copyright (c) 2020 Kiran Parajuli kiran@jankaritech.com
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU AFFERO GENERAL PUBLIC LICENSE for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public
+ * License along with this library.  If not, see <http://www.gnu.org/licenses/>.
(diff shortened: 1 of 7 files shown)


tests/acceptance/features/webUIPasswordExpire/userPasswordExpire.feature
```diff
@@ -15,3 +15,28 @@ Feature: expire user's password using the occ command
     Then the user should be redirected to a webUI page with the title "%productname%"
     When user "Alice" enters the current password, chooses a new password "newpassword" and confirms it using the webUI
     Then the user should be redirected to a webUI page with the title "Files - %productname%"
+
+  Scenario: password update with wrong old password
+    Given the administrator has enabled the days until user password expires user password policy
+    And the administrator has expired the password of user "Alice"
+    Given user "Alice" has logged in with expired password using the webUI
```
Talank, line 22:
```suggestion
    And user "Alice" has logged in with expired password using the webUI
```

tests/acceptance/features/lib/UpdatePasswordPage.php
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ * PageObject for the password policy settings Page
+ *
+ */
+class UpdatePasswordPage extends OwncloudPage {
+	/**
+	 * @var string path
+	 */
+	protected $path = '/index.php/apps/password_policy/update_password?redirect_url=/coreAgain/index.php/apps/files/';
```
Talank, line 37: Why protected? :thinking:

kiranparajuli589: IDK too...just copied and pasted it from `settingsPage` :thinking: Implemented in the same way for `$path` in other `pageObjects`.

phil-davis: behat-page-object-extension `Page.php` has:
```
abstract class Page extends DocumentElement implements PageObject
{
    /**
     * @var string|null
     */
    protected $path = null;
...
```
so this is using the same variable scope as in the parent (grandparent) that it is overriding.

phil-davis: https://www.php.net/manual/en/language.oop5.visibility.php - the parent abstract class does not want to make the value public, but expects the implementing class to be able to set, see and use it. So private also would not work. That leaves `protected`

phil-davis (approved):

PR #335, opened 2021-05-25
Add password requirement hints and Capabilities API

JanAckermann:
Add password requirement hints and Capabilities API

Fixes: https://github.com/owncloud/enterprise/issues/4576

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -39,3 +39,7 @@
 #password_policy .margin-add-top {
 	margin-top: 24px;
 }
+
+#password_policy #password_hint {
+	margin-top: 5px;
+}
--- a/lib/Capabilities.php
+++ b/lib/Capabilities.php
@@ -1,0 +1,126 @@
+<?php
+/**
+ * @author Jan Ackermann <jackermann@owncloud.com>
+ *
+ * @copyright Copyright (c) 2021, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\PasswordPolicy;
+
+use OCP\Capabilities\ICapability;
+
+/**
+ * Class Capabilities
+ *
+ * @package OCA\PasswordPolicy
+ */
+class Capabilities implements ICapability {
+
+	/** @var ConfigProvider */
+	private $configProvider;
+
+	/**
+	 * Capabilities constructor.
+	 *
+	 * @param ConfigProvider $configProvider
+	 */
+	public function __construct(ConfigProvider $configProvider) {
+		$this->configProvider = $configProvider;
+	}
+
+	/**
+	 * Return this classes capabilities
+	 *
+	 * array['password_policy']              		array Defines the capabilities
+	 *          [password_requirements]      		array Defines the password requirements
+	 *          	['minimum_characters']			integer (optional) Defines the minimum password length
+	 *              ['configuration']         		array Defines the password requirements base configuration (for e.g password managers)
+	 *              	['lower_case']       		array Defines the password requirements lower case configuration
+	 *              		['generate_from']       array Defines the possible password parameters
+	 *              			['characters']		string Defines the allowed characters (default abcdefghijklmnopqrstuvwxyz)
+	 *              		['minimum']       		integer (optional) Defines the minimum amount of lower case characters
+	 *              	['upper_case']       		array Defines the password requirements upper case configuration
+	 *              		['generate_from']       array Defines the possible password parameters
+	 *	            			['characters']      string Defines the allowed characters (default ABCDEFGHIJKLMNOPQRSTUVWXYZ)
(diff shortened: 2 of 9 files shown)


JanAckermann:
@felix-schwarz FYI

lib/Capabilities.php
```diff
@@ -0,0 +1,103 @@
...
+			],
+			'numbers' => [
+				"generate_from" => ["characters" => "1234567890"],
+			],
+			'special_characters' => [
+				"generate_from" => []
+			]],
+		];
+
+		if ($this->config->getAppValue('password_policy', 'spv_min_chars_checked') === 'on') {
```
jvillafanez, line 67: This seems dense...
Maybe you can split the data so you could loop over it, since the code is very similar
```
$data = [
  [
    'config_key_check' => 'spv_min_chars_checked',
    'config_key_value' => 'spv_min_chars_value',
    'destination' => ['minimum_characters'],
  ],
  [
    'config_key_check' => 'spv_lowercase_checked',
    'config_key_value' => 'spv_lowercase_value',
    'destination' => ['configuration','lower_case','minimum'],
  ],
];
for ($data as $item) {
  if ($this->config->getAppValue('password_policy', $item['config_key_check']) === 'on') {
    $value = \intval($this->config->getAppValue('password_policy', $item['config_key_value']));
    $target =& $generalPasswordRequirements;
    foreach ($item['destination'] as $key) {
      $target =& $target[$key];
    }
    $target = $value;
    unset($target);
  }
}
```
The problem there is that playing with references is risky. You could also try to move the assignment part to a function and pass the target array by reference
```
for ($data as $item) {
  if ($this->config->getAppValue('password_policy', $item['config_key_check']) === 'on') {
    $value = \intval($this->config->getAppValue('password_policy', $item['config_key_value']));
    $this->assignTo($generalPasswordRequirements, $item['destination'], $value);
  }
}
```

lib/Controller/PasswordController.php
```diff
@@ -189,4 +190,30 @@ public function update($current_password, $new_password, $confirm_password, $red
...
+			'spv_uppercase',
+			'spv_numbers',
+			'spv_special_chars',
+			'spv_def_special_chars'
+		];
+
+		$passwordRequirements = [];
+
+		foreach ($configKeys as $configKey) {
+			if ($this->config->getAppValue('password_policy', $configKey.'_checked') === 'on') {
```
jvillafanez, line 212: taking into account the similarities among the keys, maybe it should have its own class. It could be easier to reuse.

jvillafanez:
We might need additional info for documentation and for the clients:
* What keys will always be present?
* What keys are optional?
* Any dependency between keys? (if A is "yes" then B will be present....)
* Any default value if the admin didn't set it?

We'll also need tests to guarantee the behaviour.

lib/Capabilities.php
JanAckermann: I don't like this 'design' approach, from my POV, this makes it unnecessary complicated to read.

jvillafanez: Can we reuse the configProvider here? something like:
```
if (($spv_min_chars = $this->configProvider->get('spv_min_chars')) !== null) {
  $generalPasswordRequirements['minimum_characters'] = $spv_min_chars;
}
```

JanAckermann: For sure!

JanAckermann: @jvillafanez should the configProvider also coerce the values to the matching integer Datatype?

jvillafanez: Probably not because it can't guarantee the same type for everything. It's possible at the moment, but there is no guarantee it will remain the same.

css/styles.css
```diff
@@ -39,3 +39,7 @@
 #password_policy .margin-add-top {
 	margin-top: 24px;
 }
+
+#password_policy #password_hint {
```
JammingBen, line 43: `#password_hint` should be enough as ids are unique anyway

JammingBen (commented):
One minor nitpick from my side, LGTM overall

css/styles.css
JanAckermann: Let us stick with the syntax in the whole CSS file #password_policy ....{}

lib/Capabilities.php
```diff
@@ -0,0 +1,126 @@
...
+	 *	            			['characters']      string Defines the allowed characters (default ABCDEFGHIJKLMNOPQRSTUVWXYZ)
+	 *              		['minimum']       		integer (optional) Defines the minimum amount of upper case characters
+	 *              	['numbers']       			array Defines the password requirements numbers configuration
+	 *              		['generate_from']       array Defines the possible password parameters
+	 *          				['characters']      string Defines the allowed characters (default 1234567890)
+	 *              		['minimum']       		integer (optional) Defines the minimum amount of numbers
+	 *              	['special_characters']      array Defines the password requirements special characters configuration
+	 *              		['generate_from']       array Defines the possible password parameters
+	 *           				['characters']      string (optional) Defines the allowed characters
+	 * 							['any]				boolean (optional) Defines whether all special characters are allowed (default true)
```
jvillafanez, line 67: seems too indented

lib/ConfigProvider.php
```diff
@@ -0,0 +1,81 @@
...
+	 * @param string $configKey for example spv_min_chars without _key or _value
+	 *
+	 * @return null|string
+	 */
+	public function get(string $configKey) {
+		if ($this->config->getAppValue('password_policy', $configKey."_checked") === 'on') {
+			return $this->config->getAppValue('password_policy', $configKey."_value", null);
+		}
+
+		return  null;
```
jvillafanez, line 79: double space here

jvillafanez (approved):
tiny details, but code looks good.
We might need tests for the `configProvider->get(...)` method though

lib/Capabilities.php
JanAckermann: Thought so as well, but if you look in any editor on your local machine, it looks good

JammingBen (approved):

JanAckermann:
@jvillafanez added tests for ConfigProvider but first test is failing and I have no Idea why, any idea?

Issue #342, opened 2021-07-05
[QA] Password expire UI glitches since new login style


JanAckermann:
Actual behavior:
- Width and margin of the current password field sticks out

Expected behavior:
- Should looks nice and neat 

![image](https://user-images.githubusercontent.com/26169327/124447836-0ee04d80-dd82-11eb-99ac-836198048a90.png)

PR #343, opened 2021-07-05
Adapt ui to new login design

JanAckermann:
Fixes:
- https://github.com/owncloud/password_policy/issues/342

image

Bump min version to 10.8.0

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -53,7 +53,7 @@ config = {
 			},
 			'servers': [
 				'daily-master-qa',
-				'latest'
+				// 'latest'
 			],
 			'browsers': [
 				'chrome',
--- a/.drone.star
+++ b/.drone.star
@@ -53,7 +53,6 @@ config = {
 			},
 			'servers': [
 				'daily-master-qa',
-				// 'latest'
 			],
 			'browsers': [
 				'chrome',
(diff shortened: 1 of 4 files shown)


phil-davis:
IMO you will need to remove the test pipelines against `latest`

.drone.star
```diff
@@ -53,7 +53,7 @@ config = {
 			},
 			'servers': [
 				'daily-master-qa',
-				'latest'
+				// 'latest'
```
phil-davis, line 56: There is another "latest" further down.

And, IMO, "latest" is in the default "servers" list. So actually "daily-master-qa"-only will need to be specified in each acceptance tests section.

phil-davis: `.drone.star:56:7: got '//', want primary expression `
https://drone.owncloud.com/owncloud/password_policy/1846

`#` is a comment in starlark.

hurradieweltgehtunter (approved):

.drone.star
```diff
@@ -858,7 +856,8 @@ def acceptance(ctx):
 	errorFound = False
 
 	default = {
-		'servers': ['daily-master-qa', 'latest'],
+		#'servers': ['daily-master-qa', 'latest'],
+		'servers': ['daily-master-qa'],
```
phil-davis, line 860: I normally prefer not to change the "standard" code below `def main`. Because this code is the same in all apps, and we regularly copy the "latest and best"  code starlark changes into all the apps.

Should be OK for today. We should be bale to revert this in 1 week, when `latest` points to a 10.8 release tarball.

JanAckermann: :+1:

Issue #344, opened 2021-07-05
[QA] Password hint will be shown, even if the required length is set to 0


JanAckermann:
Steps to reproduce:
1. Enable password min lowercase rule
2. Set password min lowercase rule to 0

Actual behavior:
- Hint shows At least 0 lower case letters

Expected: behavior:
- Hint should not be displayed at all 

![image](https://user-images.githubusercontent.com/26169327/124453124-54534980-dd87-11eb-9b0b-7404c1c001d4.png)

phil-davis:
See https://github.com/owncloud/password_policy/issues/41#issuecomment-404777496

Suggestion is to prevent entering the "0" setting. If a character type requirement is "enabled" then the number must be at least 1.

JanAckermann:
@phil-davis correct, I have set the min-value of html input to 1, but this is indeed just working if you use the controls, you could easily copy and paste a 0 char. To come all over this, we need a backend validator, and this is too much for this partcular issue.

PR #345, opened 2021-07-05
Show only hints if numeric value is gt 0

JanAckermann:
Fixes: #344

### DIFF
--- a/templates/admin.php
+++ b/templates/admin.php
@@ -50,7 +50,7 @@ script('password_policy', 'ajax');
 			<li>
 				<label>
 					<input type="checkbox" name="spv_uppercase_checked" <?php if ($_['spv_uppercase_checked']): ?> checked="checked"<?php endif; ?>>
-					<input type="number" name="spv_uppercase_value" min="1 max="255" value="<?php p($_['spv_uppercase_value']) ?>"> <?php p($l->t('uppercase letters'));?>
+					<input type="number" name="spv_uppercase_value" min="1" max="255" value="<?php p($_['spv_uppercase_value']) ?>"> <?php p($l->t('uppercase letters'));?>
 				</label>
 			</li>
 			<li>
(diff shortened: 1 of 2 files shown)


phil-davis:
commit message is the wrong way around? "lt" should be "gt".
It might be worth fixing that, to avoid confusion for other devs in future.

JanAckermann:
@phil-davis thanks, holy monday !
Also adjusted min vals

templates/admin.php
```diff
@@ -38,31 +38,31 @@
...
 					<input type="checkbox" name="spv_lowercase_checked" <?php if ($_['spv_lowercase_checked']): ?> checked="checked"<?php endif; ?>>
-					<input type="number" name="spv_lowercase_value" min="0" max="255" value="<?php p($_['spv_lowercase_value']) ?>"> <?php p($l->t('lowercase letters'));?>
+					<input type="number" name="spv_lowercase_value" min="1" max="255" value="<?php p($_['spv_lowercase_value']) ?>"> <?php p($l->t('lowercase letters'));?>
 				</label>
 			</li>
 			<li>
 				<label>
 					<input type="checkbox" name="spv_uppercase_checked" <?php if ($_['spv_uppercase_checked']): ?> checked="checked"<?php endif; ?>>
-					<input type="number" name="spv_uppercase_value" min="0" max="255" value="<?php p($_['spv_uppercase_value']) ?>"> <?php p($l->t('uppercase letters'));?>
+					<input type="number" name="spv_uppercase_value" min="1 max="255" value="<?php p($_['spv_uppercase_value']) ?>"> <?php p($l->t('uppercase letters'));?>
```
jvillafanez, line 53: there is a missing double-quote after the "1"

jvillafanez:
Just to confirm, the expectation is that the value is removed in order not to use the policy, I guess. I mean, if don't require uppercase chars, the value should be removed, not just set to 0. I trust it works well with the "number" input type.

jvillafanez (approved):

JanAckermann:
@jvillafanez why this should be removed? From UX POV this makes no sense. I can deactivate a setting, and enable it later without the previous values getting lost.

image

jvillafanez:
ok, got it. Didn't see the checkboxes. :+1:

PR #385, opened 2022-11-10
[tests-only] [full-ci] Change mailhog tag to email

phil-davis:
https://github.com/owncloud/core/pull/40442 changed the email server used in tests from mailhog to inbucket. It also changed the tag used in feature files from `@mailhog` to `@email`

So make the same change here.

Issue https://github.com/owncloud/QA/issues/771

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1,6 +1,6 @@
 BANST_AWS_CLI = "banst/awscli"
 DRONE_CLI = "drone/cli:alpine"
-MAILHOG_MAILHOG = "mailhog/mailhog"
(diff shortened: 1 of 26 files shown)


phil-davis:
Fails some test scenarios related to guests. It looks like some test code needs to be updated in the guests app.

SagarGi:
Though the tests pass i have to look for the features when the email is sent to user so that it can be cleaned up as well.

SagarGi:
I looked up to the features files. All required mails seems to be pushing and cleaned up. we can merge this.

SagarGi (approved):

PR #408, opened 2024-08-14
[test-only][full-ci] fixing phan error

ghost:
related issue: https://github.com/owncloud/QA/issues/859

### DIFF
--- a/lib/HooksHandler.php
+++ b/lib/HooksHandler.php
@@ -68,16 +68,16 @@ class HooksHandler {
 	private $userNotificationConfigHandler;
 
 	public function __construct(
-		IConfig $config = null,
-		Engine $engine = null,
-		IHasher $hasher = null,
-		ITimeFactory $timeFactory = null,
-		IL10N $l10n = null,
-		PasswordExpired $passwordExpiredRule = null,
-		OldPasswordMapper $oldPasswordMapper = null,
-		ISession $session = null,
-		IManager $notificationManager = null,
-		UserNotificationConfigHandler $userNotificationConfigHandler = null
+		?IConfig $config = null,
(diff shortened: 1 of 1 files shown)


lib/HooksHandler.php
```diff
@@ -68,16 +68,16 @@ class HooksHandler {
...
-		Engine $engine = null,
-		IHasher $hasher = null,
-		ITimeFactory $timeFactory = null,
-		IL10N $l10n = null,
-		PasswordExpired $passwordExpiredRule = null,
-		OldPasswordMapper $oldPasswordMapper = null,
-		ISession $session = null,
-		IManager $notificationManager = null,
-		UserNotificationConfigHandler $userNotificationConfigHandler = null
+		?IConfig $config,
```
phil-davis, line 71:
```suggestion
		?IConfig $config = null,
```

we can leave the default `= null` so that callers are still allowed to call the constructor without specifying any parameters.

(same for other lines changed)

lib/HooksHandler.php
```diff
@@ -262,7 +262,7 @@ public function savePasswordForCreatedUser(GenericEvent $event) {
 	/**
 	 * Flags the session to require a password change
 	 */
-	protected function forcePasswordChange($firstLogin = null, IUser $user = null) {
+	protected function forcePasswordChange(?bool $firstLogin, ?IUser $user) {
```
phil-davis, line 265:
```suggestion
	protected function forcePasswordChange(?bool $firstLogin = null, ?IUser $user = null) {
```

We can leave `= null` so that the interface of this method remains the same - users cam call it without specifying parameters, if they like.

phil-davis (approved):

